PR #8, opened 2026-07-20
Make installation instructions easier for Claude

lujstn:
...does what it says on the tin!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,7 @@
 ### Claude Code
 
 ```bash
-git clone https://github.com/ayghri/i-have-adhd ./i-have-adhd
-claude plugin marketplace add ./i-have-adhd
+claude plugin marketplace add ayghri/i-have-adhd
 claude plugin install i-have-adhd@i-have-adhd
 ```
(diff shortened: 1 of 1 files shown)


isatimur (changes requested):
The repository-based Claude marketplace flow is valid. I verified add, install, list, marketplace update, plugin update, uninstall, and marketplace removal in an isolated configuration.

One documentation consistency fix is needed: this PR changes only `README.md`, while `INSTALL.md` still tells users to clone locally and later run `git pull`. I implemented and validated the matching `INSTALL.md` changes in lujstn/i-have-adhd#1: https://github.com/lujstn/i-have-adhd/pull/1.

Merge that patch into `patch-1` and this PR will be ready.

ayghri:
Thanks for this! The same fix (`claude plugin marketplace add ayghri/i-have-adhd` instead of a local clone) landed via #14, which carried an identical diff and also moved the full install docs to INSTALL.md — so main already has this. Closing as superseded. Appreciated!

PR #10, opened 2026-07-20
Add missing commands/i-have-adhd.md; document on-demand (opt-in) mode

jeremyinthebay:
Thanks for this — the ten rules are genuinely good. Two small things I hit while installing it, offered as a PR rather than an issue since they're quick.

### 1. `commands/i-have-adhd.md` is missing from the repo

Both the README ("You run `/i-have-adhd`") and INSTALL.md ("Open Claude Code, type `/i-have-adhd`") document the slash command, and the README's "How It Works" file tree lists:

```
├── commands/
│   └── i-have-adhd.md       # registers the /i-have-adhd slash command
```

But that file isn't in `main` — so on a fresh `claude plugin install`, `/i-have-adhd` never registers. This PR adds it, matching the documented behaviour (turns the rules on for the session; `stop adhd mode` / `normal mode` turns them off).

### 2. Documented an on-demand alternative

The frontmatter `description` currently says to apply the rules to *any* message "even when the user did not explicitly ask for brevity", which makes the skill always-on the moment it's installed. That sits a little oddly next to INSTALL.md's existing **Always-on (optional)** section, which implies invoking it is the default path.

Rather than change your default, I added an **On-demand only (optional)** section as the mirror image of it, with a drop-in `description` for people who'd rather call it when they want a shorter answer than have every response shaped. The ten rules are untouched — it only changes *when* they engage.

Happy to drop either half if you'd rather keep the always-on framing, or to flip the default in a follow-up if you'd prefer that instead. Either way, thanks for building it.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -151,6 +151,16 @@ npx skills remove i-have-adhd -g
 
 Manual fallback: `rm -rf ~/.cursor/skills/i-have-adhd` or `.cursor/skills/i-have-adhd`.
 
+## How activation works
+
+Three states, in order of how much you opt in:
+
+1. **Installed, not invoked.** Nothing happens. `SKILL.md` sets `disable-model-invocation: true`, so the model never sees the skill's description and never applies the rules on its own.
+2. **You type `/i-have-adhd`.** Rules on for that session. Say "stop adhd mode" or "normal mode" to turn them off.
+3. **You add the always-on config below.** Rules on from message one, every session, no command needed.
+
+There is no automatic middle ground. If you did not turn it on, it is off.
+
 ## Always-on (optional)
 
 ### Claude Code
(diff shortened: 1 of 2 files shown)


isatimur (changes requested):
The added command is a second, stale source of truth and is not required to register `/i-have-adhd`.

Current Claude Code discovers plugin skills automatically and exposes `<plugin>/skills/<name>/SKILL.md` as a slash command. Anthropic’s docs explicitly say plugin skills create `/name` shortcuts and that a same-named skill takes precedence over a command: https://code.claude.com/docs/en/slash-commands and https://code.claude.com/docs/en/plugins-reference

That means the existing `skills/i-have-adhd/SKILL.md` already provides the command. This new file copies only part of its behavior and already diverges: it mandates time estimates, omits the current ownership/safety/detail rules, and creates separate session-state semantics. Future skill fixes would have to be duplicated here.

Please remove `commands/i-have-adhd.md`. If the goal is manual-only activation, set `disable-model-invocation: true` in the single `SKILL.md` source (or document that as an explicit fork/configuration), instead of adding a parallel command. The trigger-scope problem is also being addressed centrally in #11.

ayghri:
Merged with one change: `commands/i-have-adhd.md` was dropped, since plugin skills self-register `/i-have-adhd` (per the review). The on-demand behavior shipped instead via `disable-model-invocation: true` in SKILL.md. Thanks for catching the docs contradiction.

PR #9, opened 2026-07-20
Add reproducible skill quality evaluations

isatimur:
## What changed

- add a dependency-free repository validator and CI workflow
- add 14 response-quality cases spanning correctness, autonomy, actionability, safety, and concision
- add a paired live-runner with a hard `$25` configured budget ceiling and explicit handling for unmetered providers
- add blind-judging guidance and automated release gates
- add unit tests for repository drift, case validation, and score aggregation
- add reproducible bug/feature forms, a PR checklist, contribution guidance, and a security-reporting policy

## Why

Issue #4 asks whether the skill reduces model performance. The repository currently has no reproducible way to answer that question, and token count alone cannot detect correctness, safety, or agent-autonomy regressions.

This PR establishes the measurement foundation before changing the skill instructions.

Closes no issue by itself; it is the first part of #4.

## Developer impact

Run the same checks as CI:

```bash
python3 scripts/check_repo.py
python3 scripts/run_evals.py validate
python3 -m unittest discover -s tests -v
```

See `evals/README.md` for paired live runs and blind scoring.

## Validation

- `python3 scripts/check_repo.py`
- `python3 scripts/run_evals.py validate`
- `python3 -m unittest discover -s tests -v` — 8 tests pass
- `python3 -m py_compile scripts/check_repo.py scripts/run_evals.py`
- `git diff --check`

Issue #4, opened 2026-07-19
Benchmarks on whether this reduces performance of the model


hugobiais:
This plugin honestly looks great, but I'm a bit reluctant when it comes to installing it because I'm concerned that I'll see a decrease in performance with it (since sometimes, the quality of the response of a model can be linked to the number of tokens it produces).

Would anyone be down to run benchmarks with this skill and see if we see and decline in performance or not?

Caveman does it: https://github.com/juliusbrussee/caveman

ayghri:
@hugobiais We can define how to measure the performance and I'm ready to test it out.

isatimur:
Implemented the reproducible evaluation foundation in #9:

- 14 cases across direct answers, agent autonomy, debugging, explanations, destructive actions, ambiguity, progress, output contracts, planning, and medical boundaries
- weighted correctness/autonomy/actionability/safety/concision rubric
- baseline/candidate/comparator run matrix with three trials
- hard configured budget ceiling, blind scoring, release gates, resumability, and bounded retries
- 8 unit tests plus CI

The first live Claude run produced 22 baseline responses with `$6.65` reported spend. It then stopped on the account's weekly usage limit. The partial data is preserved and resumable; no performance claim is being made until both conditions complete under the same model and rubric.

ayghri:
Update: the eval harness from #9 is now on main, trimmed to evals only. This issue stays open until someone produces baseline-vs-candidate numbers with it.

mvanhorn:
Opened #95 with a `measure` subcommand for `scripts/run_evals.py`.

It aggregates the `usage` and `cost_usd` fields the harness already records on every response row and reports per-condition token, cost and response-length totals plus a candidate-versus-baseline delta. It reuses the pairing rule from #48, so conditions are compared only when they cover identical `(case_id, trial)` sets and came from the same runner; mixing runners or partial coverage raises rather than emitting a number that looks like a measurement.

That covers the cost half of this issue. Producing the actual baseline-vs-candidate numbers is still open, since it needs a chosen model and trial count, so this references the issue rather than closing it.

TomLucidor:
Also can caveman be combined with ponytail and i-have-adhd into a superset that is easily triggered by agents?

ayghri:
> Also can caveman be combined with ponytail and i-have-adhd into a superset that is easily triggered by agents?

It should be possible. This repo will remain focused on the ADHD related rules. You can experiment with your own fork.

PR #11, opened 2026-07-20
Refine ADHD-friendly response behavior

### LINKED ISSUE #3: Suggestion: mention to use the task tools (claude) or plan tool (codex) to divide walls of texts into chunks and track them as an internal todolist
I'm working on my own (adhd-over-engineered) ADHD-friendly tooling for agentic work, and I won't be using this specific skill (but I will definitely steal some ideas), so this is not a feature request. Just sharing something I think would fit well here.

## Example

> Before we continue and decide what to do with your changes, please give me an overview. One change at a time with your explanation, 
> don't give me a wall of text. Use your todo list internal task tracking tool (and tell me how it is actually called) to track the changes and
> where we are in the explanation.

Image

I can provide more examples of prompts I've used to split walls of texts into chunks using codex's plan tool (works with claude task tools too).

## Suggestions

- the wording should be tested with codex, claude and other harnesses. I've generally had success with something like "use your internal todo list tool to track each item so we don't forget where we are" without having to mention the actual name of the tool. I don't remember if I got any friction from that but since I mostly work in codex I just mention the "plan tool" directly now.

isatimur:
## What changed

- narrow implicit activation to explicit ADHD/focus requests and substantial action-heavy work
- lead with the answer, next action, or completed result according to task ownership
- prevent the skill from delegating agent-owned edits and checks back to the user
- use native plan/task tracking for substantial work without hardcoding harness tool names
- preserve requested detail, output contracts, safety context, and rollback information
- require read-only target resolution before destructive-action confirmation
- make timing conditional and evidence-based instead of mandatory
- synchronize Claude, Codex, README, and OpenAI-facing metadata; bump Codex metadata to `0.2.0`

## Why

The original frontmatter requested activation for every message, although the README described contextual implicit invocation. Several rules also conflicted in real agent workflows: the first line always had to be a user action, timing was mandatory, and every turn had to restate state even when no workflow existed.

These conflicts could make the skill less correct, less autonomous, or more verbose—the concern raised in #4.

Closes #3.

## User impact

The skill keeps its action-first identity but now adapts to the task:

- direct questions start with answers
- agent-owned work is performed and reported
- complex work exposes progress
- detailed explanations retain necessary depth
- casual replies do not become artificial numbered workflows

Always-on behavior remains available through the documented user configuration.

## Validation

- skill quick validator: pass
- Codex plugin validator: pass
- all JSON manifests parsed with `jq`: pass
- `git diff --check`: pass
- independent forward tests: agent autonomy, destructive-action safety, and detailed OAuth explanation
- destructive-action test exposed one broad-target defect; the rule was tightened and the repeat test correctly requested the repository path and promised a read-only preview

Live paired evaluation is being tracked in #9. Claude produced 22 baseline responses with `$6.65` reported spend before the account reached its weekly usage limit; no benchmark result is claimed yet.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,14 +1,14 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "i-have-adhd",
-  "description": "Shape Claude Code output for an ADHD reader. Action-first, numbered steps, no tangents.",
+  "description": "Shape Claude Code output for an ADHD reader. Answers first, bounded steps, visible progress.",
   "owner": {
     "name": "Ayoub G."
   },
   "plugins": [
     {
       "name": "i-have-adhd",
-      "description": "Action-first, ADHD-friendly output shaping for Claude Code.",
+      "description": "Low-friction, ADHD-friendly output shaping for Claude Code.",
       "source": "./",
       "category": "productivity"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,4 +1,4 @@
 {
   "name": "i-have-adhd",
-  "description": "Shape Claude Code output for an ADHD reader: lead with the next action, number steps, suppress tangents, restate state, make wins visible."
+  "description": "Shape Claude Code output for an ADHD reader with answers first, bounded steps, visible progress, and preserved agent autonomy."
 }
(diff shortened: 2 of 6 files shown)


ayghri:
Thanks for the thorough review on #10. That's what got this resolved.

#10 is now merged with `disable-model-invocation: true` in `SKILL.md`, the mechanism you pointed to there. The description no longer reaches the model at all, so activation is fully explicit: installed and inert, `/i-have-adhd` for a session, or `CLAUDE.md` for always-on.

That covers the trigger scope question this PR set out to solve. The rest here is mostly a rewrite of the ten rules and the naming, which I'd like to keep as they are for now. Closing this one, but I appreciate the time you put in.

PR #20, opened 2026-07-21
Let the model invoke the skill by default

### LINKED ISSUE #19: `disable-model-invocation: true` makes the slash command the only way to load the skill
## Summary

`disable-model-invocation: true` makes `/i-have-adhd` the *only* way the skill can ever load. Asking Claude directly — "use the i-have-adhd skill", "apply ADHD output style" — does nothing, because the flag removes the skill's description from the model's context and blocks model invocation entirely.

I think this works against what the skill is for, and I'd like to propose removing it.

## Behaviour

From the Claude Code docs, [Control who invokes a skill](https://code.claude.com/docs/en/skills#control-who-invokes-a-skill):

| Frontmatter | You can invoke | Claude can invoke | When loaded into context |
| :-- | :-- | :-- | :-- |
| (default) | Yes | Yes | Description always in context, full skill loads when invoked |
| `disable-model-invocation: true` | Yes | **No** | **Description not in context**, full skill loads when you invoke |

Confirmed in practice: with the flag set, no amount of asking gets the skill to load. Only the slash command does.

## Why the `CLAUDE.md` route doesn't fully substitute

The README suggests `~/.claude/CLAUDE.md` as the always-on path:

```markdown
## Output style

Always follow the rules in the `i-have-adhd` skill: action-first, numbered steps, no preamble, no closers, state restated each turn.
```

That line is permanent, and it does carry real behaviour — but only the five rules it happens to name. The skill body never loads, so these never reach the model unless the user runs the slash command:

- Rule 6, specific time estimates
- Rule 7, make completed work visible
- Rule 8, matter-of-fact tone for errors
- Rule 9, cap lists at 5 items
- Rule 4, suppress tangents
- The "When to break the rules" section (explain requests, destructive actions, debug spirals, ambiguity)
- The pre-send check

So the documented always-on path delivers a subset of the skill, and the full ruleset is available only for sessions where the user remembers to type the command.

## Why this skill in particular

The docs' stated rationale for the flag is side-effecting commands:

> Use this for workflows with side effects or that you want to control timing, like `/commit`, `/deploy`, or `/send-slack-message`. You don't want Claude deciding to deploy because your code looks ready.

An output style has no side effects and no timing concerns. There is no failure mode from it applying too eagerly — worst case, an answer is more direct than needed.

More to the point: someone installs this because reading long, buried, unstructured agent output is a persistent problem for them, not a per-message one. Nobody wants action-first formatting on this reply and buried prose on the next. Requiring manual re-activation every session places the burden of remembering on precisely the users the skill exists to support, which adds exactly the kind of extra step the skill is otherwise designed to remove.

## Proposed fix

Remove the flag. This is **strictly additive** — `user-invocable` defaults to `true`, so `/i-have-adhd` keeps working exactly as it does now. The only change is that the model can also apply the skill on its own, and the description returns to context.

Anyone who prefers manual-only can add the line back in their own copy; the default seems better aimed at the common case.

PR: #20

BobbyWang0120:
Removes `disable-model-invocation: true` from `skills/i-have-adhd/SKILL.md`.

## Why

With the flag set, `/i-have-adhd` is the only way the skill can ever load. Asking Claude to apply it does nothing — per the [Claude Code docs](https://code.claude.com/docs/en/skills#control-who-invokes-a-skill), the flag blocks model invocation *and* removes the description from the model's context:

| Frontmatter | You can invoke | Claude can invoke | When loaded into context |
| :-- | :-- | :-- | :-- |
| (default) | Yes | Yes | Description always in context, full skill loads when invoked |
| `disable-model-invocation: true` | Yes | **No** | **Description not in context**, full skill loads when you invoke |

The `~/.claude/CLAUDE.md` snippet in the README keeps the five rules it names in context permanently, but the skill body never loads — so rules 4 and 6–9, the "When to break the rules" section, and the pre-send check only apply in sessions where the user remembers the command.

The docs recommend this flag for side-effecting commands (`/commit`, `/deploy`) where you don't want Claude picking the moment. An output style has neither side effects nor timing sensitivity, and users install it because buried, unstructured output is a persistent problem for them rather than a per-message one.

Full reasoning in #19.

## Impact

**Strictly additive.** `user-invocable` defaults to `true`, so `/i-have-adhd` behaves exactly as it does today. The only change is that the model can also apply the skill on its own, and the description returns to context.

Anyone preferring manual-only can add the line back locally.

## Note on ordering

Branched from `main`, independent of #17 (the YAML frontmatter fix). The two touch adjacent lines in the same block, so whichever merges second may need a one-line rebase — happy to do that whenever suits.

Fixes #19

### DIFF
--- a/skills/i-have-adhd/SKILL.md
+++ b/skills/i-have-adhd/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: i-have-adhd
 description: Shape output for a reader with ADHD: lead with the next action, number multi-step work, restate state across turns, suppress tangents, give specific time estimates, make wins visible. Invoke with /i-have-adhd; stays on until "stop adhd mode".
-disable-model-invocation: true
 ---
 
 # i-have-adhd


ayghri:
Thanks, and for #17. Declining: the design is no activation without an explicit user action. Model invocation, even with the narrowed description, is still per-turn discretion.

The gap you found is real and fixed in ffec822: the always-on block now inlines all ten rules instead of naming a skill the model can't load.

PR #21, opened 2026-07-21
Add literal-language and shortest-path rules

AdinathChaudhari:
Two concrete behaviors drawn from **W3C Cognitive Accessibility Guidance (COGA)** that the skill operationalizes everywhere else but didn't yet state explicitly:

1. **Rule 2 — shortest path, not just numbered steps.** Numbering steps helps, but COGA calls for *short critical paths*: the fewest steps to done. Added a line to cut steps the reader doesn't need and fold trivial ones in.
2. **Pre-send check — literal language.** COGA calls for literal, non-figurative wording; idioms force the reader to translate before acting. Added a scan item to replace idioms/figurative phrasing with literal wording.

Both are one-line additions to existing sections, so the "10 rules" framing and the README stay intact. Everything else in the plain-language standards (ISO 24495-1, Plain Writing Act, JAN) was already covered by the existing rules — these two were the only net-new behaviors worth adding.

### DIFF
--- a/skills/i-have-adhd/SKILL.md
+++ b/skills/i-have-adhd/SKILL.md
@@ -33,6 +33,8 @@ If the answer is a command, path, or snippet, it goes first. Prose comes after, 
 
 If the work takes more than one step, write a numbered list. Each step is one bounded action. No step contains "and then" twice.
 
+Use the fewest steps that still work. Cut any step the reader does not need, and fold trivial steps into the one before. A short path finished beats a complete path abandoned.
+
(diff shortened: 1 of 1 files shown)


ayghri:
Merged with one tweak: pre-send item 5 was compressed to match item 4's structure ("Any idiom or figurative phrase ('circle back,' 'get the ball rolling,' 'on the same page'). Replace with the literal action."). Rule count stays at 10. Thanks!

Issue #22, opened 2026-07-21
Please add Cursor compatability


gabecodez:
I use Cursor for vibe coding, and think others may find it useful if there was a skill for Cursor that produced similar behavior.

ayghri:
Cursor support already ships: `npx skills add ayghri/i-have-adhd -a cursor -y`, documented in INSTALL.md's agent-skills section. #23 added a project-level `.cursor/skills` symlink for repo cloners on top, merged as 0c79746. Closing.

PR #23, opened 2026-07-21
add Cursor skill

gabecodez:
related to my own issue, #22

I added the skill under a new Cursor skill-specific directory. Nothing was changed about the skill, but in testing may need to be adapted for Cursor-specific behavior.

### DIFF
--- a/.cursor/skills/i-have-adhd/SKILL.md
+++ b/.cursor/skills/i-have-adhd/SKILL.md
@@ -1,0 +1,1 @@
+../../../skills/i-have-adhd/SKILL.md


ayghri:
Merged with an amendment at the time: your copy was swapped for a symlink so it couldn't drift. Update: #55 showed symlinks break on Windows clones and ZIP downloads, so ccce9e7 went back to a real copy, now guarded by a CI drift check (cursor-skill-sync.yml). The shipped result is your original approach plus the sync guarantee. Thanks again!

PR #25, opened 2026-07-21
Opt-in always-on mode for Claude Code

PeterM45:
The always-on path in INSTALL.md (a CLAUDE.md snippet) only keeps the handful of rules it names in context — the full skill body never loads unless you type `/i-have-adhd` each session (#19 covers this well). For the people this skill is for, remembering a per-session command is the exact problem the skill exists to solve.

This adds a SessionStart hook that loads the whole ruleset every session, gated behind a flag file:

```
touch ~/.claude/.i-have-adhd-always   # on
rm ~/.claude/.i-have-adhd-always      # off
```

Off by default — without the flag the hook exits silently, so installing or updating changes nothing for anyone. Any failure also exits 0, so it can never block startup. Only Claude Code reads the plugin hooks field, so other harnesses keep the documented snippet approach. Same mechanism the ponytail plugin uses, so there's working precedent.

Complementary to #20: that lets the model decide to invoke the skill; this makes it deterministic for people who want it every session.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -5,5 +5,6 @@
   "author": {
     "name": "Ayoub G.",
     "url": "https://github.com/ayghri"
-  }
+  },
+  "hooks": "./hooks/hooks.json"
 }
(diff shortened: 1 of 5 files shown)


ayghri:
Merged with small amendments: rebased over the inlined always-on block, dropped the nonstandard `commandWindows` key, punctuation cleanup. Pipe-tested both ways: silent without the flag file, full frontmatter-stripped ruleset with it. Nice design: current ruleset, zero drift, still opt-in.

PR #27, opened 2026-07-21
Add Gemini CLI support

jabrahamtech:
Adds [Gemini CLI](https://github.com/google-gemini/gemini-cli) to the supported agents, alongside Claude Code, Codex, and Antigravity.

## How it works

Gemini CLI has no plugin marketplace, so this ships the two native routes — and both single-source the rules from the existing `SKILL.md` (no divergent second copy):

- **Custom command (opt-in)** — `skills/i-have-adhd/agents/gemini.toml`, copied to `~/.gemini/commands/i-have-adhd.toml` and invoked with `/i-have-adhd`. Off until you invoke it, matching this skill's default "installed, not invoked" posture. It's self-contained so it works as a global command from any directory (Gemini's `@{...}` file injection is cwd-relative, so a global command can't reference a bundled file — hence the rules are inlined here, drawn from the repo's own 10-rule summary).
- **Extension (always-on)** — `gemini-extension.json` + `GEMINI.md`, installed with `gemini extensions install https://github.com/ayghri/i-have-adhd`. `GEMINI.md` imports the full skill via `@./skills/i-have-adhd/SKILL.md`, so the rules apply from message one with no duplication.

## Changes

- **skills/i-have-adhd/agents/gemini.toml** — the opt-in command, placed next to the existing `openai.yaml` so per-agent config stays in one directory.
- **gemini-extension.json** + **GEMINI.md** — the always-on extension manifest and its imported context file.
- **README.md** — a Gemini CLI install block (command route first, extension as the one-liner alternative).
- **INSTALL.md** — Gemini CLI added to the intro and a full section (both routes, verify, update, uninstall, always-on) mirroring the other agents.

## Verified against gemini-cli docs

- Custom commands: TOML in `~/.gemini/commands/<name>.toml` → `/<name>`; `@{path}` file injection is documented as cwd-relative.
- Extensions: `gemini extensions install <github-url>` reads `gemini-extension.json`; `contextFileName` loads a context file; `GEMINI.md` supports `@./file` imports.

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -1,0 +1,5 @@
+# i-have-adhd
+
+Shape every response for a reader with ADHD. Follow the rules imported below in full: lead with the next action, number multi-step work, restate state across turns, suppress tangents, give specific time estimates, make wins visible, and cut all preamble and closers.
+
+@./skills/i-have-adhd/SKILL.md
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -311,6 +311,54 @@ Exceptions: explain fully when asked to explain. Confirm before destructive acti
 </details>
 
 <details>
+<summary><strong>Gemini CLI</strong></summary>
+
+Gemini CLI has no plugin marketplace, so there are two native routes: a **custom command** (opt-in, off until you invoke it) or an **extension** (always-on once installed). The command route matches this skill's default posture; pick it unless you want the rules on every session.
+
+### Install (command, opt-in)
+
+```bash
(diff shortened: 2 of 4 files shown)


ayghri:
Merged with amendments: README foldable dropped (repo policy), em dashes removed, and the command route's always-on subsection folded away since the extension route is the always-on path. One note: gemini.toml duplicates the ruleset by necessity (global commands must be self-contained), so it needs updating when SKILL.md changes. Thanks!

PR #29, opened 2026-07-21
Add Pi support

jabrahamtech:
Promotes [Pi](https://pi.dev) from a passing mention in the generic section to its own install section.

## Why a dedicated section

Pi implements the Agent Skills standard, so the same `skills/i-have-adhd/SKILL.md` loads directly — but two Pi-specific details the generic `npx skills` blurb omits are easy to trip on:

- **Invocation is `/skill:<name>`**, not `/<name>` like the other agents.
- It's gated behind **`"enableSkillCommands": true`** in Pi's `settings.json`.
- Skills live in `~/.pi/agent/skills/` and `.agents/skills/`.

## Changes

- **README.md** — a Pi install block (`npx skills add ... -a pi`, then `/skill:i-have-adhd`).
- **INSTALL.md** — Pi added to the intro and a full section (npx skills or filesystem copy into `~/.pi/agent/skills/`, `enableSkillCommands`, verify, update, uninstall, always-on). Removed "Pi" from the generic harness `<summary>` since it now has its own section.

## Verified against pi.dev docs

- Skill discovery dirs: `~/.pi/agent/skills/`, `~/.agents/skills/` (global); `.pi/skills/`, `.agents/skills/` (project).
- Skills register as `/skill:<name>` commands when `enableSkillCommands` is set.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -235,6 +235,82 @@ Exceptions: explain fully when asked to explain. Confirm before destructive acti
 </details>
 
 <details>
+<summary><strong>Pi</strong></summary>
+
+Pi implements the Agent Skills standard, so the same `SKILL.md` loads directly, no conversion. Pi's invocation differs from the others: skills are called as `/skill:<name>`.
+
+### Install
+
+```bash
+npx skills add ayghri/i-have-adhd -a pi -y
+```
+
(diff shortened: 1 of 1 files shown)


ayghri:
Merged with amendments: README foldable dropped (repo policy), always-on carries the full inlined ruleset. The Pi-specifics (/skill: prefix, enableSkillCommands, scan paths) were the value here; good catch that the generic section did not cover them. Thanks!

PR #30, opened 2026-07-22
Add exploration exception: don't collapse decisions to one option

jabrahamtech:
## The gap

The action-first defaults — lead with the next action (rule 1), end with **one** concrete next action (rule 3), suppress tangents (rule 4), cap lists (rule 9) — are exactly right for *"how do I fix X."* But they misfire on **exploratory / comparative** requests. Asked *"what are my options,"* *"which library,"* or *"how should I approach this,"* the skill collapses the answer into a single, possibly-wrong path and hides the decision space. For a reader who came to *choose*, that's not brevity — it's being railroaded into someone else's pick.

## The fix

One new case in the existing **"When to break the rules"** section (no rule rewrites, no frontmatter changes):

> **5. Exploration, or a choice between approaches.** When the reader wants to weigh options … do NOT collapse to one path. Give a bounded, ranked menu: 2 to 4 real options, one line each with its main trade-off, recommendation first ("Recommended: X — because…"). Here the options ARE the answer …

## Why it's consistent, not a loosening

It follows the skill's **own** core premise. The skill opens with "working memory is small — anything not on screen is forgotten." For a decision, the *alternatives* are precisely what must stay on screen; collapsing to one forces the reader either to hold "but what were the other choices?" in memory (which the skill says they can't) or to accept a hidden default. The menu is still ADHD-friendly — capped at 2–4, one line each, ranked, recommendation-first, no preamble. It just distinguishes *"how do I fix this"* (one action) from *"what should I do"* (a short ranked menu).

Scoped deliberately small (one bullet), in the spirit of #21 — not the broad behavior rewrite of the closed #11.

### DIFF
--- a/skills/i-have-adhd/SKILL.md
+++ b/skills/i-have-adhd/SKILL.md
@@ -114,6 +114,7 @@ Override the defaults when:
 2. Destructive action ahead (`rm -rf`, force push, schema migration, dropping a table). Confirm before acting. Safety wins over brevity.
 3. Debug spiral. If the last three turns have been "still broken," stop iterating on code. Name the assumption that might be wrong. Ask one diagnostic question.
 4. Real ambiguity in the request. One short clarifying question beats guessing and rewriting.
+5. A rule fights the task. When a rule would delete the answer itself, the task wins; the shape stays. Example: "what are my options" gets 2 to 4 ranked options with one-line trade-offs, recommendation first, not one path. The options are the answer.
 
 ## Pre-send check
 


ayghri:
Nice idea, and something I've been grappling with myself: when does the ruleset stop serving the task. Merged with a rephrase that generalizes it. Case 5 is now "A rule fights the task", with your exploration menu as the worked example (2 to 4 ranked options, one-line trade-offs, recommendation first). The "shape stays" clause keeps it from becoming a license for preamble. Thanks!

PR #36, opened 2026-07-22
fix(skill): instruct persistence and define the "stop adhd mode" off-switch

### LINKED ISSUE #31: SKILL.md never tells the model to persist, and never defines "stop adhd mode"
## Summary

The frontmatter promises session persistence and an off-switch:

```yaml
description: 'Shape output for a reader with ADHD: … Invoke with /i-have-adhd; stays on until "stop adhd mode".'
```

`INSTALL.md:197` repeats it:

> **You type `/i-have-adhd`.** Rules on for that session. "stop adhd mode" or "normal mode" turns them off.

But the body of `SKILL.md` — the part the model actually reads — contains neither instruction:

```
$ grep -n 'stop adhd mode\|normal mode\|every response\|persist' skills/i-have-adhd/SKILL.md
3:description: '… stays on until "stop adhd mode".'
```

One hit, in the frontmatter.

## Why it matters

`description` is selection metadata, not a behavioural instruction the model is told to obey. The skill body is injected on invocation and says nothing about later turns, so:

1. **Persistence is not actually instructed.** The rules apply on the invocation turn, then drift back toward default style as the conversation grows and the injected block loses salience. "Rules on for that session" is documented but never implemented.
2. **The off-switch is undefined.** The model was never told that "stop adhd mode" or "normal mode" means anything. Whether it honours the documented phrase is luck.

Output-style skills that stay on across turns generally carry an explicit persistence block ("applies to every response … off only on `stop X`"). This one has none.

## Suggested fix

Add a short `## Persistence` section directly after the intro line, before `## What ADHD changes about reading`, stating that the rules apply to every response for the rest of the session and that "stop adhd mode" / "normal mode" turns them off.

## Verify

```bash
grep -n 'stop adhd mode\|normal mode' skills/i-have-adhd/SKILL.md
```

should match in the body, not only in the frontmatter. Behaviourally: invoke `/i-have-adhd`, run ~10 unrelated turns, confirm the reply shape holds; then say "stop adhd mode" and confirm it drops.

Happy to send a PR.

thejesh23:
Fixes #31

## Problem

The frontmatter promises `Invoke with /i-have-adhd; stays on until "stop adhd mode"`, and `INSTALL.md:197` repeats it. The body of `SKILL.md` — the part the model actually reads — says neither thing:

```
$ grep -n 'stop adhd mode\|normal mode\|every response\|persist' skills/i-have-adhd/SKILL.md
3:description: '… stays on until "stop adhd mode".'
```

One hit, in the frontmatter. `description` is selection metadata, not an instruction the model is told to obey, so persistence was never actually instructed and "stop adhd mode" was undefined behaviour.

## Change

Six lines: a `## Persistence` section placed directly after the intro, before the ADHD facts, so it is read before the rules it governs. It states that the rules hold for the rest of the session and that "stop adhd mode" / "normal mode" turns them off.

No rule text changed; nothing removed.

## Verify

```bash
grep -n 'stop adhd mode\|normal mode' skills/i-have-adhd/SKILL.md
```

now matches in the body, not only the frontmatter. Behaviourally: invoke `/i-have-adhd`, run ~10 unrelated turns, confirm the reply shape holds; then say "stop adhd mode" and confirm it drops.

Frontmatter re-parsed as YAML after the edit — unchanged and still valid (`name`, `description`, `disable-model-invocation`).

## Conflicts

Touches only `skills/i-have-adhd/SKILL.md`, in a region no open PR edits (#20 edits the frontmatter, #21/#26/#30 edit the rule list further down).

### DIFF
--- a/skills/i-have-adhd/SKILL.md
+++ b/skills/i-have-adhd/SKILL.md
@@ -8,6 +8,12 @@ disable-model-invocation: true
 
 The reader has ADHD. Output is not just brief. It is shaped so an ADHD brain can act on it.
 
+## Persistence
+
+These rules apply to every response for the rest of the session, not only this one. They do not expire after a few turns and they do not lapse when the topic changes. If you are unsure whether they still apply, they do.
+
+Turn them off only when the reader says "stop adhd mode" or "normal mode". Confirm in one line, then return to your default style.
+
 ## What ADHD changes about reading
 
 Five facts drive every rule below:


ayghri:
Merged. This closed a real gap: with `disable-model-invocation` the description never reaches the model, so the body was the only place the persistence contract and off-switch could live. Thanks.

PR #40, opened 2026-07-22
fix(antigravity): drop the dead `$schema` URL from root `plugin.json`

### LINKED ISSUE #35: Root `plugin.json` points at a non-existent Antigravity schema URL
## Summary

`plugin.json:2`:

```json
"$schema": "https://antigravity.google/schemas/v1/plugin.json"
```

```
$ curl -s -o /dev/null -w "%{http_code}\n" -L https://antigravity.google/schemas/v1/plugin.json
404
```

The host is live (`https://antigravity.google/` → 200); the schema path is not. Neighbouring guesses 404 as well:

```
https://antigravity.google/schemas/plugin.json          -> 404
https://antigravity.google/schema/v1/plugin.json        -> 404
https://antigravity.google/schemas/v1/plugin.schema.json-> 404
```

The Antigravity plugin docs (https://antigravity.google/docs/plugins) document the manifest as a marker file and never reference a JSON schema:

> Every plugin must have a `plugin.json` file at its root. This file identifies the directory as a plugin.
> ```json
> { "name": "my-custom-plugin" }
> ```
> The `name` field is optional and defaults to the directory name if omitted.

## Why it matters

Same class as #34: the key is editor-only, so nothing fails at load time — but an editor that honours `$schema` tries to fetch it, fails, and the manifest ends up with no validation while looking like it has some. A dead URL is worse than no URL, because it reads as authoritative.

## Suggested fix

Drop the key until Antigravity publishes a schema:

```diff
 {
-  "$schema": "https://antigravity.google/schemas/v1/plugin.json",
   "name": "i-have-adhd",
   "description": "…"
 }
```

If there is a canonical Antigravity schema URL I could not find, I am glad to point at that instead — just say which.

## Verify

```bash
python3 -c "import json;json.load(open('plugin.json'))"   # still valid JSON
```

and Antigravity still loads the plugin (`name` is all the documented manifest requires).

thejesh23:
Fixes #35

## Problem

`plugin.json:2` points at a schema that does not exist:

```
$ curl -s -o /dev/null -w "%{http_code}\n" -L https://antigravity.google/schemas/v1/plugin.json
404
```

The host itself is live (`https://antigravity.google/` → 200); only the schema path is missing. Nearby guesses 404 too (`/schemas/plugin.json`, `/schema/v1/plugin.json`, `/schemas/v1/plugin.schema.json`), and the Antigravity plugin docs never mention a schema — they describe the manifest as a marker file:

> Every plugin must have a `plugin.json` file at its root. This file identifies the directory as a plugin.
> ```json
> { "name": "my-custom-plugin" }
> ```
> The `name` field is optional and defaults to the directory name if omitted.

— https://antigravity.google/docs/plugins

## Change

Delete the one key. Editor-only metadata, so nothing about loading changes — but a dead URL is worse than none: the editor tries to fetch it, fails, and the manifest ends up unvalidated while looking authoritative.

If there is a canonical Antigravity schema URL I could not find, I am happy to point at that instead — just say which and I will amend.

## Verify

```bash
python3 -c "import json;print(list(json.load(open('plugin.json')).keys()))"
# ['name', 'description']
```

Still valid JSON, and `name` is all the documented Antigravity manifest requires.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -1,5 +1,4 @@
 {
-  "$schema": "https://antigravity.google/schemas/v1/plugin.json",
   "name": "i-have-adhd",
   "description": "Shape Antigravity output for an ADHD reader: lead with the next action, number steps, suppress tangents, restate state, make wins visible."
 }


ayghri:
Merged. Verified the schema URL 404s; dropping it was right. Thanks.

PR #41, opened 2026-07-22
feat(claude): native output style for always-on (full ruleset)

shihyuho:
## What

This ships the i-have-adhd ruleset as a native Claude Code output style. Always-on then needs no `CLAUDE.md` edit, and it loads the full ruleset in every session.

You turn it on once and it stays on:

```json
// .claude/settings.json  (or ~/.claude/settings.json for every project)
{ "outputStyle": "i-have-adhd" }
```

Or open `/config`, pick Output style, and choose `i-have-adhd`. To turn it off, switch back to Default.

## Why

The documented always-on path pastes a prose snippet into `~/.claude/CLAUDE.md`. As #19 and #25 both note, that snippet only carries the few rules it names. The full skill body never loads, so rules 4 and 6 to 9, "When to break the rules", and the pre-send check silently don't apply.

Output styles are Claude Code's built-in way to reshape how it replies, which is what this plugin does. That is where this belongs, and it brings the whole ruleset with it.

## Relation to #25

#25 closes the same always-on gap with a `SessionStart` hook and a flag file. This PR is a different mechanism aimed at the same goal:

| | This PR (output style) | #25 (SessionStart hook) |
|---|---|---|
| Mechanism | Native Claude Code primitive | Custom `hooks/always-on.js` + `hooks.json` |
| Turn on | `/config` picker, or one `outputStyle` line | `touch ~/.claude/.i-have-adhd-always` |
| Turn off | Switch to Default (or any other style) in `/config` | `rm ~/.claude/.i-have-adhd-always` |
| Full ruleset loaded | Yes, the whole body is the style | Yes, it injects `SKILL.md` |
| Discoverable | Listed in `/config` | A flag file you have to know exists |
| Extra runtime | None | A JS process at every SessionStart |

Turning it on and off is the same picker you already use to switch between Default, Explanatory, and the rest, so there is no separate dotfile to remember. I don't think it has to be one or the other. Happy to defer to whichever you prefer, or to close this if #25 is the direction you want.

## Design

- `keep-coding-instructions: true` keeps Claude Code's engineering behavior and adds the ADHD shaping on top. Without it, the style would replace the coding system prompt.
- `SKILL.md` stays the single source of truth. `scripts/build-output-style.sh` regenerates the style from it: a frontmatter swap, with the body copied verbatim. A CI check (`output-style-sync.yml`) regenerates and fails the build if the two drift, so they can't fall out of sync.
- Still opt-in. There is no `force-for-plugin`, so installing changes nothing until you pick it. "If you didn't turn it on, it's off" still holds.
- Scoped to Claude Code. Only `output-styles/` is read here. The Codex, Antigravity, and generic-harness paths are untouched.

## Files

- `output-styles/i-have-adhd.md`: the output style (generated, full ruleset).
- `scripts/build-output-style.sh`: regenerates it from `SKILL.md`.
- `.github/workflows/output-style-sync.yml`: CI check that fails if the two drift.
- `INSTALL.md`: the Claude Code "Always-on" section now documents the output-style path.

## Verify

```
/config -> Output style -> i-have-adhd    # appears, selectable, persists across sessions
```

### DIFF
--- a/.github/workflows/output-style-sync.yml
+++ b/.github/workflows/output-style-sync.yml
@@ -1,0 +1,24 @@
+name: output-style in sync
+
+# Fails if output-styles/i-have-adhd.md is stale, i.e. SKILL.md was edited
+# without rerunning scripts/build-output-style.sh. Keeps SKILL.md the single
+# source of truth.
+on:
+  pull_request:
+    paths:
+      - skills/i-have-adhd/SKILL.md
+      - output-styles/**
+      - scripts/build-output-style.sh
+      - .github/workflows/output-style-sync.yml
+  push:
+    branches: [main]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Regenerate the output style
+        run: sh scripts/build-output-style.sh
+      - name: Fail if it differs from the committed file
+        run: git diff --exit-code output-styles/i-have-adhd.md
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,13 +37,17 @@ Or keep it installed and turn it off: `claude plugin disable i-have-adhd`.
 
 ### Always-on (optional)
 
-Add to `~/.claude/CLAUDE.md`:
+The plugin bundles an `i-have-adhd` output style. Select it once and it persists across sessions — no `CLAUDE.md` edit needed.
 
-```markdown
-## Output style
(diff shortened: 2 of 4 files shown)


ayghri:
Thanks for the writeup, the comparison table made this an easy call. Passing on it, but only because #25 landed first and I want exactly one always-on mechanism for Claude Code. Yours is honestly the cleaner design, so if the hook ever becomes a pain this is what I'd replace it with. I'd happily take the build-script + CI drift check on its own though, we have other derived copies (gemini.toml) that could use it.

PR #48, opened 2026-07-22
fix(evals): reject unpaired conditions before scoring the release gate

### LINKED ISSUE #44: Release gate passes when baseline and candidate were judged on different cases
## Summary

`scripts/run_evals.py` `summarize_scores()` averages each condition independently and never checks that the conditions cover the same `(case_id, trial)` rows. A comparison across mismatched sets produces a confident, wrong verdict.

## Repro

Baseline judged on the two hardest cases, candidate judged on the easiest one:

```jsonl
{"case_id":"destructive-action","trial":1,"condition":"baseline","correctness":2,"autonomy":2,"actionability":2,"safety":2,"concision":2,"blocker":false,"notes":""}
{"case_id":"medical-boundary","trial":1,"condition":"baseline","correctness":2,"autonomy":2,"actionability":2,"safety":2,"concision":2,"blocker":false,"notes":""}
{"case_id":"direct-answer","trial":1,"condition":"candidate","correctness":5,"autonomy":5,"actionability":5,"safety":5,"concision":5,"blocker":false,"notes":""}
```

```
$ python3 scripts/run_evals.py score scores.jsonl
  "baseline":  { "rows": 2, "weighted_score": 2.0 },
  "candidate": { "rows": 1, "weighted_score": 5.0 },
  "release_gate": { "passed": true, "reasons": [] }
```

`rows: 2` vs `rows: 1` is the only hint, and nothing acts on it.

## Why it matters

This gate is the ship / do-not-ship decision for a change to the ruleset. `evals/README.md:51` already states the requirement —

> Do not compare conditions produced with different cases, models, trial counts, or rubrics.

— but it is enforced only by the operator remembering. A partial candidate run (provider failure part-way, a `--case` subset, a judge who skipped rows) silently yields "passed". The same gap lets a duplicated score row be counted twice and pull an average.

## Suggested fix

In `summarize_scores`, after the baseline/candidate presence check, compare the `(case_id, trial)` multiset of every condition against baseline and raise with the concrete difference (missing / unmatched rows), plus reject duplicate `(case_id, trial)` rows inside one condition.

## Verify

The repro above raises instead of returning `passed: true`; the existing 6 tests still pass (`python3 -m unittest discover -s tests`).

Happy to send a PR with a regression test.

thejesh23:
Fixes #44

## Problem

`summarize_scores()` averaged each condition independently and never checked the conditions covered the same rows. Baseline judged on the two hardest cases, candidate on the easiest one:

```
$ python3 scripts/run_evals.py score scores.jsonl
  "baseline":  { "rows": 2, "weighted_score": 2.0 },
  "candidate": { "rows": 1, "weighted_score": 5.0 },
  "release_gate": { "passed": true, "reasons": [] }
```

`evals/README.md:51` already states the rule — *"Do not compare conditions produced with different cases, models, trial counts, or rubrics"* — but only the operator enforced it. A candidate run cut short by a provider failure, a `--case` subset, or a judge who skipped rows all yield a confident "passed". Duplicate score rows had the same effect on the averages.

## Change

`_check_pairing()` compares the `(case_id, trial)` multiset of every condition against baseline and raises with the concrete difference; duplicates within a condition are rejected first. Called from `summarize_scores` right after the existing baseline/candidate presence check. Two regression tests added.

## Verify

```
$ python3 -m unittest discover -s tests -q
Ran 8 tests in 0.001s
OK
```
(6 existing + 2 new; the suite was green before the change too)

The repro above now fails loudly instead of passing:
```
ValueError: candidate was not judged on the same rows as baseline: missing destructive-action/trial 1, medical-boundary/trial 1; unmatched direct-answer/trial 1
```

A properly paired file is unaffected — 2 baseline + 2 candidate rows over the same cases still returns `{"passed": true, "reasons": []}`.

## Conflicts

Touches `summarize_scores` and the test file only. My other two `run_evals.py` PRs sit in `_condition_prompt` (#…) and the `run` loop, so the three merge in any order.

### DIFF
--- a/scripts/run_evals.py
+++ b/scripts/run_evals.py
@@ -9,7 +9,7 @@ import shlex
 import subprocess
 import sys
 import time
-from collections import defaultdict
+from collections import Counter, defaultdict
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


ayghri:
Merged as-is. Verified the gate really did compare unpaired rows before this; the Counter-based coverage check plus the two tests close it properly. All 8 tests pass on main.

PR #49, opened 2026-07-22
fix(evals): fail when `--case` matches no evaluation case

### LINKED ISSUE #45: `--case` with an unknown id runs nothing and exits 0
## Summary

`scripts/run_evals.py:200` filters cases with `if args.case and case["id"] not in args.case: continue`. An id that matches nothing filters out every case, and the run reports success.

## Repro

```
$ python3 scripts/run_evals.py run --runner claude --condition baseline \
    --output evals/results/responses.jsonl --trials 1 --case direct-anwser
Reported cost: $0.0000
$ echo $?
0
```

No rows written, no warning, exit 0. `Reported cost: $0.0000` reads like a run that legitimately cost nothing — resume already skips completed rows, so a zero-cost run is a normal thing to see.

## Why it matters

Typos in a `--case` id are easy (`destructive-action`, `medical-boundary`, `multi-step-progress`), and the failure is invisible: the operator believes that condition was measured. Feed the resulting file to `score` and it compares whatever happens to be in there.

## Suggested fix

Validate `--case` values against the loaded catalog before the loop and fail with the unknown ids listed.

## Verify

```
$ python3 scripts/run_evals.py run ... --case direct-anwser
ValueError: --case matched no evaluation case: direct-anwser
```
and a correct `--case direct-answer` still runs.

thejesh23:
Fixes #45

## Problem

`run_evals.py:200` filters with `if args.case and case["id"] not in args.case: continue`, so an id that matches nothing filters out every case and the run reports success:

```
$ python3 scripts/run_evals.py run --runner claude --condition baseline \
    --output evals/results/responses.jsonl --trials 1 --case direct-anwser
Reported cost: $0.0000
$ echo $?
0
```

Zero rows, no warning. `$0.0000` is not a red flag either — resume skips completed rows, so a zero-cost run is normal. The operator believes that condition was measured; `score` then compares whatever is in the file.

## Change

Three lines in `run_evaluations`, after the existing case validation:

```python
unknown = sorted(set(args.case or []) - {case["id"] for case in cases})
if unknown:
    raise ValueError(f"--case matched no evaluation case: {", ".join(unknown)}")
```

## Verify

```
$ python3 scripts/run_evals.py run ... --case direct-anwser
ValueError: --case matched no evaluation case: direct-anwser
$ echo $?
1
```

Unaffected paths, checked against a stub runner: `--case direct-answer` runs that one case; omitting `--case` runs all 14. `python3 -m unittest discover -s tests -q` gives 6 tests, OK.

## Conflicts

Only the `run_evaluations` preamble. My other two `run_evals.py` PRs touch `summarize_scores` (#48) and `_condition_prompt`, so all three merge in any order.

### DIFF
--- a/scripts/run_evals.py
+++ b/scripts/run_evals.py
@@ -177,6 +177,9 @@ def run_evaluations(args: argparse.Namespace) -> int:
     errors = validate_cases(cases)
     if errors:
         raise ValueError("\n".join(errors))
+    unknown = sorted(set(args.case or []) - {case["id"] for case in cases})
+    if unknown:
+        raise ValueError(f"--case matched no evaluation case: {', '.join(unknown)}")
     config = json.loads(args.runner_config.read_text(encoding="utf-8"))
     runner = config[args.runner]
     command = list(runner["command"])


ayghri:
Merged as-is. Confirmed the silent exit-0 at the filter in run_evaluations; the guard is the right fix.

PR #56, opened 2026-07-22
fix(evals): isolate the claude runner from the operator's own config

### LINKED ISSUE #52: Claude eval runner inherits the operator's global config — the always-on flag injects the skill under test into the baseline condition
## Summary

The `claude` command in `evals/runners.example.json` does not isolate the eval call from the operator's user-level Claude Code configuration. Plugins, `SessionStart`/`UserPromptSubmit` hooks, memory, and `CLAUDE.md` all load into every condition — including `baseline`.

The `codex` runner in the same file already isolates (`--ephemeral --ignore-user-config`); the claude runner has no equivalent flag.

## Live repro

I ran the documented baseline command on a machine with an output-style plugin installed (caveman-mode, a SessionStart hook). The harness's own `baseline` row for `direct-answer` ("What is 17 multiplied by 6?") came back:

```
"response": "102.\n\nSide note: caveman statusline badge not configured. Want me add statusLine…"
"cost_usd": 0.14687
```

Style-shaped answer plus a tangent from my hook — in the **baseline** condition. Same prompt with isolation added:

```
$ claude --disable-slash-commands --print --output-format json --no-session-persistence \
    --tools "" --setting-sources "" --max-budget-usd 0.5 "What is 17 multiplied by 6?"
"result": "17 × 6 = 102"
"total_cost_usd": 0.0319
```

Clean, and 4.6× cheaper — the operator config was also inflating every call's token bill.

## Why it matters

The worst case is this repo's own recommended setup: `INSTALL.md` tells users to `touch ~/.claude/.i-have-adhd-always`, which makes the plugin's SessionStart hook inject the **full i-have-adhd ruleset into every session — including every eval subprocess**. An operator with that flag set runs a "baseline" that already contains the skill under test. Baseline ≈ candidate, the weighted scores converge, and the release gate fails with "Candidate weighted score did not beat baseline" — or worse, published numbers are silently invalid. The people most likely to run these evals are exactly the people with the flag set.

## Suggested fix

Add `--setting-sources ""` to the claude command in `evals/runners.example.json` (verified working with OAuth; `result` and `total_cost_usd` unaffected), and a sentence in `evals/README.md` saying both runners isolate operator config and why that must be preserved.

## Verify

Run the baseline command on a machine with any always-on output style configured: without the flag the styling leaks into responses; with it the response is clean.

Happy to send the PR.

thejesh23:
Fixes #52

## Problem

The `claude` command in `evals/runners.example.json` inherits the operator's user-level Claude Code configuration — plugins, `SessionStart`/`UserPromptSubmit` hooks, memory, `CLAUDE.md`, **and the saved model/effort settings** — into every condition, baseline included. The `codex` runner in the same file already isolates (`--ephemeral --ignore-user-config`); the claude one had no equivalent.

Live repro on a machine with an output-style plugin installed — the harness's own `baseline` row for `direct-answer` ("What is 17 multiplied by 6?"):

```
"response": "102.\n\nSide note: caveman statusline badge not configured. Want me add statusLine…"
```

Operator styling plus an operator-hook tangent, in the baseline. Sharpest case: `INSTALL.md` recommends `touch ~/.claude/.i-have-adhd-always`, whose SessionStart hook injects the **full i-have-adhd ruleset into every eval subprocess** — an operator with the flag set runs a baseline that already contains the skill under test, so the comparison measures the skill against itself and the gate false-fails on "weighted score did not beat baseline".

The model leak is its own hazard: pre-fix, the eval silently ran whatever model the operator had configured — in the repro, the contaminated call ran the operator's Fable 5 while the isolated call fell back to the CLI default Opus 4.8 (verified to the cent from `usage` under each model's pricing). Different operators would publish numbers measured on different models without knowing it.

## Change

- `evals/runners.example.json`: add `--setting-sources ""` (config isolation) **and** `--model claude-opus-4-8` (isolation drops the saved model setting, so the runner must pin one — otherwise the model varies by operator and CLI release).
- `evals/README.md`: paragraph on why both the isolation and the pin must be kept, with the always-on-flag case named.

## Verify

Through the harness itself, same machine, same case:

| runner config | response | cache-write tokens | billed model |
|---|---|---|---|
| before | `102.\n\nSide note: caveman statusline badge…` (221 output tokens) | 6,790 | operator's configured model (Fable 5) |
| after (isolated + pinned) | `102` (3 output tokens) | 3,221 | `claude-opus-4-8`, cost matches Opus pricing to the cent |

Context contamination alone (same-model comparison) roughly halves the input tokens and eliminates the styled output. OAuth unaffected; `result` and `total_cost_usd` still present (budget accounting and the unmetered gate depend on them). `python3 -c "import json; json.load(open('evals/runners.example.json'))"` passes.

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -32,7 +32,9 @@ python3 scripts/run_evals.py run \
 
 The default Claude runner reports dollar cost and receives the remaining condition budget on every call. Runners without cost reporting are rejected unless `--allow-unmetered` is supplied; use that flag only when the provider account has its own hard cap.
 
-Both example runners isolate the call from the operator's own agent configuration — `--setting-sources ""` for Claude, `--ignore-user-config --ephemeral` for Codex. Keep that isolation when adding runners: without it, user-level plugins, hooks, memory, and output styles leak into every condition and shape the responses being judged. The sharpest case is this repo's own always-on flag (`~/.claude/.i-have-adhd-always`), which would inject the full i-have-adhd ruleset into the **baseline** condition and make the comparison measure the skill against itself.
+Both example runners isolate the call from the operator's own agent configuration: `--setting-sources ""` for Claude, `--ignore-user-config --ephemeral` for Codex. Keep that isolation when adding runners: without it, user-level plugins, hooks, memory, and output styles leak into every condition and shape the responses being judged. The sharpest case is this repo's own always-on flag (`~/.claude/.i-have-adhd-always`), which would inject the full i-have-adhd ruleset into the **baseline** condition and make the comparison measure the skill against itself.
+
+Isolation also drops the operator's saved model and effort settings, so the claude runner pins `--model` explicitly. Keep a pin when editing the runner: without one, the eval silently runs whatever the operator (or the CLI release) defaults to; the model would vary between operators and over time, and per-token cost varies with it. The pinned model is part of the result: record it with published numbers, as below.
 
 Runs are resumable: rerun the same command after a provider failure and completed `(case, trial, condition, runner)` rows are skipped. Each incomplete call is retried twice by default, and the final provider error is preserved.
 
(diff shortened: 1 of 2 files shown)


ayghri:
Merged with small amendments: em dashes out, and the hardcoded per-model pricing dropped from the README since published numbers rot. The baseline-contamination catch was the sharpest one yet; the always-on flag from #25 really would have measured the skill against itself. Thanks.

PR #57, opened 2026-07-22
fix(evals): reject unmetered runners before the first paid call

### LINKED ISSUE #53: Unmetered-runner rejection is raised after the first paid call, though it is knowable before any call
## Summary

`scripts/run_evals.py:279` raises

```
Runner did not report dollar cost; rerun with --allow-unmetered only when the provider has a separate hard spending cap.
```

**after** the subprocess has already run. But for the `text` and `codex-jsonl` response formats, `_parse_response` returns `cost = None` unconditionally (`:187`, `:205`) — no output those runners produce can ever satisfy the check. Whether the run is unmetered is fully determined by the runner config before the first call.

## Impact

First run with the shipped codex runner and no `--allow-unmetered`: one full API call is made and paid for, then the harness raises and discards the response — the row is never written, so the rerun pays for that call again. One wasted paid call per configuration mistake, guaranteed on the most likely first mistake.

## Suggested fix

Pre-flight in `run_evaluations`, before the loop:

```python
if response_format != "claude-json" and not args.allow_unmetered:
    raise RuntimeError(...)  # same message
```

Keep the existing per-call check as a backstop for `claude-json` responses that omit `total_cost_usd` (that case is only knowable at runtime).

## Verify

A `text`-format runner without `--allow-unmetered` now fails before executing anything (no subprocess, no output file); with the flag it runs; the claude runner is unaffected. Existing tests stay green.

thejesh23:
Fixes #53

## Problem

`run_evals.py` raised `Runner did not report dollar cost; rerun with --allow-unmetered…` **after** the subprocess had run. For the `text` and `codex-jsonl` formats, `_parse_response` returns `cost = None` unconditionally — no output those runners can produce ever passes the check — so the failure is fully determined by config before any call. First codex run without `--allow-unmetered`: one full API call paid, response discarded (row never written), rerun pays for it again.

## Change

Pre-flight in `run_evaluations`, right after `response_format` resolves:

```python
if response_format != "claude-json" and not args.allow_unmetered:
    raise RuntimeError(
        f"The {response_format!r} response format never reports dollar cost; rerun with "
        "--allow-unmetered only when the provider has a separate hard spending cap."
    )
```

The per-call check stays — for `claude-json` responses that omit `total_cost_usd`, which is only knowable at runtime.

Regression test included: a `text` stub runner whose command touches a marker file — without `--allow-unmetered` the run raises with the marker **absent** and no output file created (nothing executed, nothing paid); flipping the flag runs it and the marker appears.

## Verify

```
$ python3 -m unittest discover -s tests
Ran 9 tests … OK
```

Manual: a `codex-jsonl` config without the flag now fails before executing anything; the shipped claude runner is unaffected (verified against a claude-json stub).

## Conflicts

`run_evaluations` preamble + test file. Independent of #56 (evals config/README only).

### DIFF
--- a/scripts/run_evals.py
+++ b/scripts/run_evals.py
@@ -218,6 +218,11 @@ def run_evaluations(args: argparse.Namespace) -> int:
     runner = config[args.runner]
     command = list(runner["command"])
     response_format = runner.get("response_format", "text")
+    if response_format != "claude-json" and not args.allow_unmetered:
+        raise RuntimeError(
+            f"The {response_format!r} response format never reports dollar cost; rerun with "
+            "--allow-unmetered only when the provider has a separate hard spending cap."
+        )
     reported_cost = 0.0
     prior_rows = read_jsonl(args.output) if args.output.exists() else []
     done = completed_keys(prior_rows)
--- a/tests/test_run_evals.py
+++ b/tests/test_run_evals.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import sys
 import tempfile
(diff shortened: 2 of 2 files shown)


ayghri:
Merged as-is. The stub-runner test proving no invocation before rejection is exactly how this should be locked in. Thanks.

PR #66, opened 2026-07-22
hooks: make always-on SessionStart hook portable (drop node dependency)

awscullinan:
The SessionStart hook invoked 'node "${CLAUDE_PLUGIN_ROOT}/hooks/always-on.js"',
which fails with 'node: command not found' on any machine where Node is not on
PATH (e.g. a stock Linux box or a Bun/uv-only setup). Claude Code does not ship
or expose a runtime that hook scripts can rely on, so a hook that hard-depends
on 'node' is not portable.

The script only reads a flag file and prints SKILL.md with its YAML frontmatter
stripped -- no reason to need Node. Reimplement it as a POSIX sh script
(always-on.sh) that runs in the shell Claude Code already uses for command hooks
(sh on macOS/Linux, Git Bash on Windows). Behavior is unchanged: honors
$CLAUDE_CONFIG_DIR, only fires when the flag exists, and exits 0 on any failure
so it never blocks session start.


ayghri:
Merged with one amendment: the script now resolves SKILL.md relative to $0 instead of trusting $CLAUDE_PLUGIN_ROOT to be exported into the hook's environment; with the env var unset the original exited silently, the same invisible-failure shape as #61. Tested flag/no-flag/env-unset locally, and plugin-load-check passed. Good catch on the node dependency, thanks.

PR #67, opened 2026-07-22
docs: add Chinese and Japanese README translations

tura-ai-agent:
## Summary

- Add complete Simplified Chinese and Japanese README translations.
- Add language navigation links to all three README files.

## Validation

- Verified matching line, heading, code fence, list, table, link, and inline-code counts.
- Verified all relative links resolve.
- `python -m unittest discover -s tests -v`: 8 passed; 1 existing Windows-specific marker-file test failed.


ayghri:
@claude review this PR. Beyond correctness, check against the house rules:

- README covers Claude Code and Codex only; every other agent lives in INSTALL.md as one self-contained foldable (install, verify, update, uninstall, always-on), no cross-references between foldables.
- `skills/i-have-adhd/SKILL.md` is the single source of truth. Any copy of the ruleset (always-on blocks, gemini.toml, the .cursor copy, translations) must be updated in the same PR and flagged as derived.
- Always-on blocks must inline the full ruleset, never "follow the rules in the skill": with `disable-model-invocation` the model cannot load the skill from a pointer.
- Nothing activates without explicit user action. `disable-model-invocation` is Claude Code only, so claims about other harnesses must state what actually happens there, and untested harness behavior must be labeled untested.
- Keep the 10-rules framing: new behavior goes into existing sections (break-the-rules cases, pre-send items), in the file's terse voice with Bad/Good examples.
- Frontmatter must parse under a strict YAML parser and the plugin must still load (CI: plugin-load-check).
- No em dashes, no filler, no closing pleasantries. Translations keep the original structure and rule numbering and state which commit of SKILL.md they translate; the wording is free to be idiomatic in the target language, do not enforce English-specific phrasing.

ayghri:
Merged, thanks! Skipping the review's commit-reference suggestion on purpose: translations are assumed to track the same commit as the README they mirror, and any drift gets handled like a normal docs fix.

PR #71, opened 2026-07-25
Fix Windows always-on SessionStart hook: launch via Node instead of sh

### LINKED ISSUE #70: SessionStart always-on hook errors on Windows: 'sh' and ${CLAUDE_PLUGIN_ROOT} don't resolve under PowerShell
### Summary

On Windows, the always-on `SessionStart` hook errors at every session start. Claude Code executes hook `command` strings via **PowerShell** (the default Windows shell), and the shipped command relies on two POSIX assumptions that don't hold there:

```json
"command": "sh \"${CLAUDE_PLUGIN_ROOT}/hooks/always-on.sh\""
```

1. **`sh` is not on PATH** under PowerShell (Git Bash's `sh` isn't globally available) → *command not found*.
2. **`${CLAUDE_PLUGIN_ROOT}` is not expanded** by PowerShell — `${...}` is PowerShell variable syntax, so it resolves to an empty string (env vars are `$env:CLAUDE_PLUGIN_ROOT`). The command effectively becomes `sh "/hooks/always-on.sh"`.

The hook therefore exits non-zero and Claude Code reports a startup hook error on **every** launch. The `always-on.sh` logic itself is fine (and correctly exits 0) — it just never gets to run.

This is distinct from #61 / #62 (duplicate hook *registration*, where the hook never loads). Here the hook loads and executes, but the *command syntax* isn't Windows-portable. Still present on `main`.

### Environment

- Windows 11, Claude Code, plugin v0.1.0
- Hook runner on Windows: PowerShell

### Reproduction

1. Windows + Claude Code, plugin installed, `touch ~/.claude/.i-have-adhd-always`.
2. Start a session → startup hook error.

Root cause confirmed from PowerShell:

```powershell
PS> Get-Command sh            # -> not found
PS> "${CLAUDE_PLUGIN_ROOT}"   # -> empty string (PS var syntax, not the env var)
```

### Expected

On Windows the always-on hook should run and inject the ruleset (or no-op silently) with no startup error.

### Suggested fix

Stop depending on a POSIX `sh` plus outer-shell variable expansion. The most portable option is to launch via **Node**, which Claude Code always ships and which is OS-agnostic:

```json
"command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/always-on.mjs\""
```

…porting the ~20 lines of `always-on.sh` to a small `.mjs` (read the flag file, read `SKILL.md`, strip frontmatter, print). Alternatively, provide OS-specific hook entries (a PowerShell command on Windows).

### Workaround for other Windows users meanwhile

Repoint the hook at a PowerShell reimplementation of `always-on.sh`:

```json
"command": "powershell -NoProfile -ExecutionPolicy Bypass -File \"<abs-path>\\i-have-adhd-always.ps1\""
```

where the `.ps1` mirrors the script: flag-file check -> read newest `SKILL.md` -> strip YAML frontmatter -> print the banner + body (exit 0 with no output otherwise).

Souptik96:
Fixes #70

## Root cause (confirmed on Windows + Claude Code 2.1.218)
The SessionStart hook ran `sh "${CLAUDE_PLUGIN_ROOT}/hooks/always-on.sh"`. On Windows,
Claude Code runs hook commands through PowerShell, where `sh` is not on PATH — so every
session start reported a hook error.

I instrumented the hook and verified how Claude Code actually handles the command on Windows:

```
ARGV=["…node.exe","C:/Users/…/i-have-adhd/"]
```

That is, **Claude Code expands `${CLAUDE_PLUGIN_ROOT}` itself** (to a real, forward-slash
path — not empty) *before* the command reaches the shell, and `node <path>` executes fine.
So the "`${CLAUDE_PLUGIN_ROOT}` resolves to empty under PowerShell" theory isn't what breaks
it on current Claude Code — the only blocker is the missing `sh`. (#66 had moved the hook
from Node to `sh` to drop the Node dependency, which is what introduced the Windows breakage.)

## Fix
- Add `hooks/always-on.mjs` — a Node port of `always-on.sh` with identical behavior: opt-in
  flag check (`$CLAUDE_CONFIG_DIR/.i-have-adhd-always`), resolve `SKILL.md` relative to the
  script (`import.meta.url`, not a trusted env var), strip YAML frontmatter, print the banner
  + body, swallow all errors and exit 0 (never blocks session start).
- Point `hooks/hooks.json` at `node "${CLAUDE_PLUGIN_ROOT}/hooks/always-on.mjs"`. Node ships
  with Claude Code on all platforms and runs identically, so this works under PowerShell/cmd
  (Windows) and sh (macOS/Linux). No reliance on a POSIX shell or on the shell interpolating
  `${VAR}` (Claude Code does the substitution).
- Keep `always-on.sh` unchanged as the POSIX fallback.

A combined `node … || sh …` command was intentionally avoided: Windows PowerShell 5.1 doesn't
support `||` (would reintroduce a parse error) and Windows has no `sh` regardless.

## Testing (Windows 11, PowerShell, Claude Code 2.1.218, Node 20)
- `node hooks/always-on.mjs` with the flag **absent** → no output, exit 0 (silent no-op preserved).
- `node hooks/always-on.mjs` with the flag **present** → prints `ADHD MODE ACTIVE (always-on)…`
  followed by the frontmatter-stripped `SKILL.md` body, exit 0.
- End-to-end: installed the plugin from a local marketplace, created the flag, ran a Claude
  Code session → the SessionStart hook injected the ruleset (model confirmed receipt) with
  **no startup hook error**.
- Not verified locally: the POSIX `sh` fallback (no `sh`/Git-Bash on the test box). `always-on.sh`
  is unchanged.

### DIFF
--- a/.github/workflows/plugin-load-check.yml
+++ b/.github/workflows/plugin-load-check.yml
@@ -9,11 +9,25 @@ on:
       - .claude-plugin/**
       - hooks/**
       - skills/**
+      - tests/test_always_on_hooks.py
       - .github/workflows/plugin-load-check.yml
   push:
     branches: [main]
 
 jobs:
+  hook-parity:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+      - name: Test native hook implementations
+        run: python -m unittest tests.test_always_on_hooks -v
+
   load:
     runs-on: ubuntu-latest
     steps:
--- a/hooks/always-on.mjs
+++ b/hooks/always-on.mjs
@@ -1,0 +1,40 @@
+// SessionStart hook: injects the full i-have-adhd ruleset when the user has
+// opted in by creating $CLAUDE_CONFIG_DIR/.i-have-adhd-always (default ~/.claude).
+// Never blocks session start: any failure exits 0.
+//
+// Runs under Node so it works on macOS, Linux, and Windows without depending on
+// a POSIX shell (`sh`) being on PATH. Claude Code ships a Node runtime and
+// expands ${CLAUDE_PLUGIN_ROOT} in the hook command before invoking it, so the
+// launcher works under PowerShell (Windows) as well as sh (macOS/Linux).
+
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+try {
+  const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), ".claude");
+  const flagPath = path.join(claudeDir, ".i-have-adhd-always");
+
+  // Only fire when the user has opted in.
+  if (!fs.existsSync(flagPath)) process.exit(0);
+
+  // Resolve SKILL.md relative to this script's own location, not a trusted env var.
+  const scriptDir = path.dirname(fileURLToPath(import.meta.url));
+  const skillPath = path.join(scriptDir, "..", "skills", "i-have-adhd", "SKILL.md");
+  if (!fs.existsSync(skillPath)) process.exit(0);
+
+  // Strip a leading YAML frontmatter block (--- ... --- at the very top of file).
+  const body = fs
+    .readFileSync(skillPath, "utf8")
+    .replace(/^---\r?\n[\s\S]*?\r?\n---\r?\n/, "");
+
+  process.stdout.write(
+    "ADHD MODE ACTIVE (always-on). The ruleset below applies to every response. " +
+      '"stop adhd mode" turns it off for this session; ' +
+      `delete ${flagPath} to turn always-on off for good.\n\n${body}\n`,
+  );
+} catch {
+  // Never block session start.
+  process.exit(0);
+}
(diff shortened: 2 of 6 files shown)


Souptik96:
Thanks for the hardening commit @ayghri - the exec-form `hooks.json` is the documented right call (the hooks reference recommends exactly `"command": "node", "args": [...]` for path placeholders, since exec form skips shell parsing entirely), and the parity test matrix is a big improvement.

Ran your commit on real Windows PowerShell 5.1: all 3 tests pass. While probing edges I found one remaining parity gap and pushed `795181b` for it:

**Unclosed frontmatter** (`---` opener, no closing delimiter): Node's regex only strips a complete fence so it kept the whole file, but both fallbacks entered frontmatter mode and never left - the user got the banner promising "the ruleset below applies" followed by nothing. The fix treats an unterminated fence as not-frontmatter in all three runtimes, with a parity test pinning identical output.

Verified on Windows: 4/4 pass (node + powershell runtimes; sh probed separately via Git bash awk 5.4 - unclosed keeps the body, and your trailing-whitespace case still strips correctly). Negative control: reverting only `always-on.ps1` fails exactly the new test with the banner-plus-empty-body output.

Souptik96:
Thanks for the merge and for the hardening pass @ayghri - the exec-form config and the cross-OS test matrix made this much stronger than the original PR.

One timing note: my follow-up commit `795181b` (unclosed-frontmatter parity fix + its test, described in the comment above) landed on the branch a few minutes after you merged, so it isn't in `main` - the sh/PowerShell fallbacks there still emit an empty ruleset for a SKILL.md whose `---` fence never closes. It's a small, already-tested diff; happy to open a tiny follow-up PR if you'd like it.

ayghri:
Then you might need to start a new PR for the added changes.

PR #78, opened 2026-07-28
Add native session-persistent Pi mode

nishuzumi:
## Summary

- add a native Pi extension with `/i-have-adhd`, `--adhd`, session persistence, and an `ADHD ON` footer status
- preserve `/skill:i-have-adhd` as an alias without injecting the rules twice
- support an always-on marker under Pi's agent configuration directory
- update the Pi installation and lifecycle documentation to use Pi's package manager
- add an isolated RPC smoke check for package loading, commands, state restoration, input pass-through, and footer status

## Behavior

Installing the repository remains opt-in. The extension does not change model behavior until the mode is invoked, restored from the current session, enabled with `--adhd`, or enabled by the marker file. A saved session choice takes precedence over the flag and marker defaults.

## Verification

- `python3 scripts/check_pi_extension.py`
- `python3 -m unittest discover -s tests -v`
- `python3 scripts/run_evals.py validate`


peachest:
Nice work — the compaction re-injection and disable notice are solid.

One gap: the diff doesn't include a root `package.json` with the `"pi": { "extensions": [...] }` field. Pi uses this to discover the extension entry point. Without it, `pi install` adds the package but the extension doesn't load (I hit this exact issue on my own attempt). See how ponytail and pi-insight declare it:

```json
"pi": {
  "extensions": ["./extensions/i-have-adhd.ts"],
  "skills": ["./skills"]
}
```

PR #86, opened 2026-07-29
Add rule 11 — append the cut reasoning as a compressed Full picture block

vicmanb22:
Title: Add rule 11 — append the cut reasoning as a compressed Full picture block

---

## What

A eleventh rule, plus the three carve-outs it needs to survive the existing ones.

Rules 1–10 strip reasoning along with the padding. For a lookup or a one-command fix that is
exactly right. For a judgment call it is lossy: the reader gets an action with no trade-off, no
rejected alternative, and no failure mode — so they cannot tell a considered recommendation from
a guess, and the skill quietly costs them the ability to disagree with it.

Rule 11 gives that back without touching the shape above the line. At the very bottom, under a
`---` rule and a bold `**Full picture**` label:

- Under 120 words, or five bullets.
- Ordered: load-bearing reason → what was rejected → what would break it.
- Never a new action. Actions live above the line, always.
- Nothing already stated above.
- **Skipped entirely** when the answer carried no judgment. A block that only says "this is the
  standard way to do it" is noise.

The first line is still the action. A reader who stops at the `---` loses nothing.

## Why the carve-outs

Rule 10 forbids recaps, and pre-send-check items 2 and 3 delete trailing recaps and "by the way"
sidebars. All three would have deleted the new block. Rather than a blanket exemption, each carve-out
is drawn on the real distinction:

- A recap restates what the reader just read; the block adds what the action-shaped answer had to
  leave out.
- A sidebar interrupts the answer; the block sits after it and is skippable by design.

And a new closing test keeps it honest: *could the reader skip the block entirely and still act
correctly?* If no, something load-bearing drifted below the line — move it back up. That test is
what stops rule 11 from becoming a hiding place for the answer.

## Files

| File | Change |
| --- | --- |
| `skills/i-have-adhd/SKILL.md` | Rule 11, the two carve-outs, the closing test, description clause |
| `.cursor/skills/i-have-adhd/SKILL.md` | Kept byte-identical to the above |
| `README.md` + 4 translations | Rule count 10 → 11, one-line entry appended |
| `skills/i-have-adhd/agents/gemini.toml` | Prompt is self-contained, so rule 11 is spelled out there too |
| `evals/cases.jsonl` | Two cases, both directions |

`python3 scripts/run_evals.py validate` passes.

## Eval cases

Deliberately two, because the interesting failure is over-application:

- `judgment-context` — a worktree-vs-branch question. Should commit to one option, say why it beats
  the rejected one, and name a condition that would make it wrong.
- `no-judgment-no-context` — "what git command shows the current branch". Should give the command
  and add no trade-off commentary at all.

Criteria are worded by substance, not by format, so they do not leak the condition to a blind judge
or auto-fail baseline on a label it was never told to emit.

## Notes for review

- **The translations are mine and are one-line summaries** (ja / ko / vi / zh-CN). Worth a native
  read before merge — the full rule text stays English in `SKILL.md`, so the exposure is one line
  each.
- **No version bump included.** `plugin.json` / `.codex-plugin/plugin.json` still say `0.1.0`; that
  felt like the maintainer's call rather than the contributor's.
- Rule 11 has been running as a local fork since 2026-07-29. The failure mode it was written from:
  the skill worked, and then a run of judgment calls arrived as bare imperatives with nothing to
  push back on.


vicmanb22:
Withdrawing this one — I'm keeping it as a personal adaptation in my own fork rather than putting it in your queue, since it's a judgment call about how much reasoning belongs in an action-first format and that's yours to make, not mine.

Rule 11 and its carve-outs live at https://github.com/vicmanb22/i-have-adhd if they're ever of interest. Thanks for the skill — it's genuinely changed how I read agent output.

PR #94, opened 2026-08-01
feat: pi always-on hook via extension

ykyyyy:
## What

Pi has no hooks.json like Claude Code, so the `SessionStart` hook has no direct equivalent — the extension system is the mechanism. This PR adds `extensions/i-have-adhd-always.ts`, a pi extension that mirrors `hooks/always-on.sh`:

- When `$PI_CODING_AGENT_DIR/.i-have-adhd-always` (default `~/.pi/agent/.i-have-adhd-always`) exists, it appends the full `SKILL.md` ruleset to the system prompt of every turn, so the rules apply from message one in every session — same UX as `touch ~/.claude/.i-have-adhd-always`.
- Pure no-op without the flag; installing the extension changes nothing by itself.
- Resolves `SKILL.md` from the repo layout (in-place install) or `~/.pi/agent/skills/` (the `npx skills add` install), strips frontmatter like `always-on.sh` does.
- `"stop adhd mode"` still turns it off for the current session.

Also updates INSTALL.md (Pi section + "How activation works").

## Why the system-prompt hook, not message injection

`before_agent_start`'s `message` return re-injects every turn (context pollution); `sessionManager.appendMessage` on the live session is undocumented/read-only-by-docs. Returning `systemPrompt` from `before_agent_start` is the documented, idempotent pattern (pi's own `claude-rules.ts`/`pirate.ts` examples) and survives startup/resume/reload/fork identically.

## Verified

Ran headless (`pi -p`) with a scratch `PI_CODING_AGENT_DIR`:
- flag present → ruleset injected into the system prompt, frontmatter stripped
- flag absent → no injection

## Notes

- No CI change: the repo's workflows only cover the Claude plugin paths.
- The condensed `AGENTS.md` snippet stays in the Pi section as the no-extension fallback.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -285,7 +285,24 @@ Or delete `~/.pi/agent/skills/i-have-adhd`.
 
 ### Always-on (optional)
 
-Add to your project `AGENTS.md`:
+Pi has no hooks.json, so the Claude `SessionStart` hook has no direct equivalent — but an extension behaves the same way: it loads the full ruleset at the start of every session, no `/skill:i-have-adhd` needed. Install the extension once, then toggle it with a flag file:
+
+```bash
+cp extensions/i-have-adhd-always.ts ~/.pi/agent/extensions/
+touch ~/.pi/agent/.i-have-adhd-always     # always-on for every session
+```
+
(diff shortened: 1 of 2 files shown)


ykyyyy:
Closing — author wants to review before submission. Will reopen after review.

PR #95, opened 2026-08-01
feat(evals): report token and cost deltas between conditions

mvanhorn:
## Summary

Adds `python3 scripts/run_evals.py measure <responses.jsonl>` to compare recorded generation token usage, cost, and response length across paired conditions. This makes it possible to see whether the skill adds input overhead or cost even when the answer becomes shorter, alongside the existing quality results.

Refs #4. This is reporting infrastructure, not a new live benchmark result or a claim that the skill increases or decreases spending.

## Maintainer amendments

- Updated against current main, preserving the contributor's commits and the existing prompt/frontmatter handling.
- Added input-token deltas alongside output-token, cost, and response-length deltas. Positive means increased usage/cost.
- Reject invalid costs/counts, overflowing cost totals, duplicate or unpaired rows, mixed runners, and conflicting or partially supplied model metadata. Missing measurements remain null, and percentages against zero remain null.
- Reuse the existing pairing check directly, removing the proposed refactor of score validation. Existing scoring, generation, response parsing, and prompting functions are unchanged.
- Document what recorded costs include and the limits of older metadata. Scope remains three files: the runner module, its tests, and eval documentation; no root README or skill changes.

## Verification

- `python3 -m unittest discover -s tests -v` — 70 passed, without provider calls. Includes both supported usage shapes, cache accounting, missing versus zero measurements, invalid costs/tokens, duplicate/incomplete pairs, model metadata mismatches, and the actual `measure` CLI.
- CLI fixture: +50% input tokens, -50% output tokens, +50% cost; expected deltas reported and input file unchanged. These are synthetic test values, not skill benchmark results.
- `python3 scripts/run_evals.py validate` — passed.
- AST comparison against main — `_check_pairing`, `summarize_scores`, `run_evaluations`, `_parse_response`, and `_condition_prompt` unchanged.
- `git diff --check` — passed.

## Limits and side effects

Read-only aggregation: no network, provider calls, or writes to response files. It measures recorded generation rows; judge costs and unrecorded failed/retried calls are excluded.

Matching runner aliases do not establish identical model/configuration. If all rows omit model metadata, the summary returns `model: null`; verify original model/CLI settings before interpreting deltas. When metadata exists, every row must report the same model. Scenario captures currently omit token counts, and their stored response length includes transcript formatting and user prompts.

No new live evaluation was run or needed to test this aggregation command.

## Provenance

Original implementation by @mvanhorn; original agent/model provenance was not supplied. Maintainer amendments are hybrid: owner-directed review and Codex (GPT-6) implementation, testing, and documentation. Final human review and merge remain pending.

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -38,6 +38,16 @@ Isolation also drops the operator's saved model and effort settings, so the clau
 
 Runs are resumable: rerun the same command after a provider failure and completed `(case, trial, condition, runner)` rows are skipped. Each incomplete call is retried twice by default, and the final provider error is preserved.
 
+## Measure
+
+Aggregate token usage, reported cost, and response length from the completed responses file:
+
+```bash
+python3 scripts/run_evals.py measure evals/results/responses.jsonl
+```
+
+The command refuses to compare conditions produced by different runners or conditions with unequal `(case_id, trial)` coverage. This is the same comparability rule the release gate applies to judged scores.
+
 ## Judge and score
 
 Blind the `condition` field before judging. Write one JSON object per response with these fields:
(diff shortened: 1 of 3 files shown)


mvanhorn:
Thanks for the merge, @ayghri. The token and cost deltas should make it easier to tell when a skill trades a shorter answer for more input overhead.

PR #103, opened 2026-08-06
hooks: keep the skill body when frontmatter is never closed

Souptik96:
Follow-up to #71, as requested in https://github.com/ayghri/i-have-adhd/pull/71#issuecomment-5205760730 — this is the commit that raced the merge (`795181b`), cherry-picked cleanly onto current `main`.

## Problem

The three hook runtimes disagree on a malformed `SKILL.md` whose opening `---` fence is never closed:

- **Node** (`always-on.mjs`): the regex only strips a *complete* fence, so it keeps the whole file — reasonable.
- **sh + PowerShell fallbacks**: enter frontmatter mode on line 1 and never leave. The user gets the banner promising *"The ruleset below applies to every response"* followed by **nothing**.

That breaks the parity invariant the #71 test matrix asserts, just on a different input.

## Fix

Treat an unterminated fence as not-frontmatter in all three runtimes:

- `always-on.ps1`: only strip when the closing delimiter was actually found (drop the `$bodyStart = $lines.Length` presumption).
- `always-on.sh`: two-pass awk — first pass decides whether the fence ever closes, second strips only if it did.
- `tests/test_always_on_hooks.py`: new parity case pinning all runtimes to identical output for this input, alongside the existing trailing-whitespace case.

## Verification (Windows 11)

- 4/4 tests pass under PowerShell 5.1 (node + powershell runtimes; `sh` is not on PATH for `shutil.which` here, so the sh hook was probed separately through Git bash awk 5.4 — unclosed fence keeps the body, and the #71 trailing-whitespace case still strips correctly).
- Negative control: reverting only `always-on.ps1` fails exactly the new test, with the banner-plus-empty-body output verbatim.
- Cherry-pick onto current `main` (`9a7e1f5`) applied with no conflicts; suite re-run green on this base.

AI assistance: drafted with Kiro CLI; all changes reviewed and verified by me.

### DIFF
--- a/hooks/always-on.ps1
+++ b/hooks/always-on.ps1
@@ -25,7 +25,8 @@ try {
   $bodyStart = 0
 
   if ($lines.Length -gt 0 -and $lines[0] -match '^---\s*$') {
-    $bodyStart = $lines.Length
+    # Only treat the block as frontmatter when the closing delimiter exists;
+    # an unterminated fence is not frontmatter, so keep the whole file.
     for ($i = 1; $i -lt $lines.Length; $i++) {
       if ($lines[$i] -match '^---\s*$') {
         $bodyStart = $i + 1
--- a/hooks/always-on.sh
+++ b/hooks/always-on.sh
@@ -19,11 +19,19 @@ skill_path="$script_dir/../skills/i-have-adhd/SKILL.md"
 [ -f "$skill_path" ] || exit 0
 
 # Strip a leading YAML frontmatter block (--- ... --- at the very top of file).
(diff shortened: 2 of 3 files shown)


ayghri:
While defensive coding is good, for our case we control "SKILL.md" and can adjust it as needed.

Future note: move frontmatter check to validation tests and keep the hooks simple.

Thanks @Souptik96

Souptik96:
Thanks for merging, and for the future direction on validation vs. hooks - agreed that's the cleaner split long-term. Appreciate you taking the time on both of these.

PR #105, opened 2026-08-07
docs: Kimi Code plugin's install instruction added

commie70:
docs(kimi): Add kimi-code plugin's multi-lingual instructions.


ayghri:
I tried to put the install instructions in alphabetical order, so I expect Kimi to come after Gemini, but I'm not sure I propagated this to all the markdowns.

commie70:
Now Qwen takes the third place, ahead of Codex. Maybe Kimi should be after Hermes by alphabetical order? I would like to check and to correct all the markdowns and open another PR.

PR #107, opened 2026-08-10
fix(hooks): run SessionStart correctly in Codex

Y0lan:
## Problem

Codex command hooks use the `command` field and pass the hook payload as JSON on stdin. The current hook stores the script path in `args`, so Codex launches bare `node`. Node then parses the `SessionStart` JSON as JavaScript and exits with code 1:

```text
SyntaxError: Unexpected token ':'
```

This produces `SessionStart hook (failed)` on every Codex session start, even when the always-on flag is absent.

## Fix

- Launch the module through one `node -e` command understood by both Codex and Claude Code.
- Resolve the installed plugin path inside Node from `CLAUDE_PLUGIN_ROOT` or `PLUGIN_ROOT`, avoiding platform-specific shell variable syntax.
- Preserve the existing opt-in behavior and the native shell/PowerShell fallbacks.

Codex documents plugin-root environment variables and a single command string for command hooks: https://learn.chatgpt.com/codex/hooks

## Validation

- Added a regression test that sends a real-shaped `SessionStart` JSON object on stdin.
- Covered both flag-present and flag-absent behavior with a plugin path containing spaces.
- `python3 -m unittest discover -s tests -v`: 15 tests pass.
- `claude plugin validate .`: validation passed.

### DIFF
--- a/hooks/always-on.mjs
+++ b/hooks/always-on.mjs
@@ -2,9 +2,9 @@
 // opted in by creating $CLAUDE_CONFIG_DIR/.i-have-adhd-always (default ~/.claude).
 // Never blocks session start: any failure exits 0.
 //
-// Runs under Node so it works on macOS, Linux, and Windows without depending on
-// a POSIX shell (`sh`) being on PATH. The hook uses exec form, so Claude Code
-// passes the script path directly without PowerShell or POSIX-shell parsing.
+// Runs under Node so it works on macOS, Linux, and Windows. The shared Claude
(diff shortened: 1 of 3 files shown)


lwu-st:
Following this, hope you get a chance to merge it soon @ayghri -- thank you

ayghri:
Check the review comments by Copilot.

Personally, I'm familiar with python more than node, and I prefer handling the env variables within the script itself, unless `node -e` is the only way to expose them...

PR #109, opened 2026-08-10
docs: clarify task list overflow handling

### LINKED ISSUE #96: Misinterpreted "Cap lists at 5 items"
Based on my experience the limitation can be followed too literally (literal 5 limit) and/or applied on lists it shouldn't be applied on (lists that influence further results and shouldn't be really limited, grouped, etc at that point as they might be relevant later on).

In multiple occasions it didn't do what IMO the original intention was. It doesn't just split a list - that is bigger than 5 items - into "do now" vs "later," or "must" vs "nice to have.". It just literally "capped the list at 5 items". It should be called more like "split up", "create sections", "group",  etc.

Also on top of this not every question has 5 (or less) valid answer, and there are lists that must have all the actually relevant items, even if it means having more than 5. I totally agree that the endless lists some newer models produce are bad, but intentionally dropping very relevant findings aren't really better either - and with the current wording it did happen.

For my environment I completely removed that section, and it still brings most of the benefit - with better accuracy - so that is an option as well, but there might be less-nuclear options than completely removing that section:
* the wording should be tweaked to drop anything barely resembling "limiting", and probably explicitly scoped to influence only the final output
* the item count limit should be made dynamic to be adjusted by the LLM based on the context/task itself - or at least increasesed to 10 so cuts less by accident

My point is that - if I get it right - this skill should be about influencing the output the user gets, not the internal thought process. Sure burning tokens on something that won't be shown/used later on is a waste, but I doubt we can influence the internal workings of a thought process without affecting result/finding validity/completeness, so instructions that might be interpreted as such should be made less ambigous and more focused about the scope.

I wanted to talk about this before I do any PR - or you might take it yourself.

https://github.com/ayghri/i-have-adhd/blob/d05af1e4ac2259846e81686d14180d46d84acc2d/skills/i-have-adhd/SKILL.md?plain=1#L105-L108

SulimanAbdulrazzaq:
﻿## Summary
- clarify that the five-item cap applies to task lists
- retain relevant overflow items in reserve instead of discarding them
- describe how to surface the reserve and handle exhaustive-list requests

## Validation
- `python -m pytest -q` (12 passed; 1 existing Windows shell-runner failure because `sh` is unavailable)
- `git diff --check`

Fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ A skill for your coding assistant that stops it from burying the answer. Action 
 6. Specific time estimates (minutes, not "a bit").
 7. Make wins visible.
 8. Matter-of-fact errors.
-9. Cap lists at 5 items.
+9. Cap task lists at 5 items; retain relevant items in reserve.
 10. No preamble. No recap. No closers.
 
 ## Tune it
--- a/skills/i-have-adhd/SKILL.md
+++ b/skills/i-have-adhd/SKILL.md
@@ -100,9 +100,9 @@ Never use "Uh oh," "Oh no," or "There seems to be a problem." State cause and fi
 Bad: "Uh oh, the test is failing. There seems to be an issue..."
 Good: "Test fails at `auth.spec.ts:42`: expected 200, got 401. Cause: missing auth header. Fix: add `Authorization: Bearer ${token}` to the request."
 
-### 9. Cap lists at 5 items
+### 9. Cap task lists at 5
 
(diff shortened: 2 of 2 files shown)


ayghri:
There is an ongoing discussion in #96 about this

SulimanAbdulrazzaq:
Thanks for pointing me to #96. The Rule 9 wording that landed through #149 and your follow-up commits covers what this PR set out to do: long lists are grouped and ranked, and extra relevant items are retained instead of dropped. This PR now conflicts with that wording, so I'm closing it.

PR #115, opened 2026-08-14
Support OMP session context API

kevinkavi2001:
## Summary

Adds native Oh My Pi (OMP) compatibility while preserving the existing Pi integration.

OMP exposes session messages through `buildSessionContext().messages`, while Pi uses `buildContextEntries()`. The extension previously called the Pi-only API directly, causing an OMP
startup error and preventing the ADHD rules from being restored from session history.

## Changes

- Add a compatibility adapter for both session-context APIs:
  - OMP: `buildSessionContext().messages`
  - Pi: `buildContextEntries()`
- Preserve active/disabled marker ordering for both message shapes.
- Keep the existing Pi API as a fallback.
- Declare the extension through the `omp` package manifest field.
- Add neutral OMP installation and update instructions.
- Add focused Pi/OMP compatibility and package regression checks.
- Make the evaluation-runner test portable on Windows by removing its dependency on `sh` and `touch`.

## Validation

```text
bun scripts/check_context_compat.ts
Pi/OMP context compatibility checks passed
 ```

 ```text
python -B -m unittest discover -s tests -v
Ran 15 tests
OK
 ```

 ```text
python -B scripts/check_pi_extension.py --runtime omp
omp extension smoke test passed
 ```

 An OMP TUI smoke test also confirmed:

 1. Launching OMP with `--adhd` starts the extension with `ADHD ON`.
 2. /i-have-adhd toggles the mode successfully.
 3. No buildContextEntries is not a function error occurs.

 ## Compatibility notes

 - Existing Pi behavior and buildContextEntries() support remain intact.
 - No skill rules or user-facing output behavior were changed.
 - No upstream package version bump is included.
 - The default on-demand state remains disabled unless `--adhd`, the always-on flag, or saved session state enables it.
 - The Pi executable was not available for a direct runtime smoke test in the development environment; its legacy session-entry shape is covered by the compatibility regression check.

### DIFF
--- a/extensions/context-compat.ts
+++ b/extensions/context-compat.ts
@@ -12,35 +12,30 @@ type CompatibleSessionManager = {
 export function contextMessages(
   sessionManager: unknown,
 ): readonly ContextMessageMarker[] {
+  // Context inspection is advisory. If a runtime changes its session-manager
+  // API or temporarily cannot build context, report "not present" so the
+  // caller can safely re-inject the rules instead of breaking session startup.
   if (sessionManager === null || typeof sessionManager !== "object") {
-    throw new Error("Unsupported session manager: expected an object");
+    return [];
   }
 
   const compatible = sessionManager as CompatibleSessionManager;
 
-  if (typeof compatible.buildSessionContext === "function") {
-    const messages = compatible.buildSessionContext().messages;
-    if (!Array.isArray(messages)) {
-      throw new Error(
-        "Unsupported session manager: buildSessionContext() returned no messages array",
-      );
+  try {
+    if (typeof compatible.buildSessionContext === "function") {
+      const messages = compatible.buildSessionContext().messages;
+      return Array.isArray(messages) ? messages : [];
     }
-    return messages;
+
+    if (typeof compatible.buildContextEntries === "function") {
+      const entries = compatible.buildContextEntries();
+      return Array.isArray(entries) ? entries : [];
+    }
+  } catch {
+    return [];
   }
 
-  if (typeof compatible.buildContextEntries === "function") {
-    const entries = compatible.buildContextEntries();
-    if (!Array.isArray(entries)) {
-      throw new Error(
-        "Unsupported session manager: buildContextEntries() returned no entries array",
-      );
-    }
-    return entries;
-  }
-
-  throw new Error(
-    "Unsupported session manager: expected buildSessionContext() or buildContextEntries()",
-  );
+  return [];
 }
 
 export function latestMarkerIsActive(
--- a/extensions/context-compat.ts
+++ b/extensions/context-compat.ts
@@ -1,0 +1,66 @@
+export type ContextMessageMarker = Readonly<{
+  type?: string;
+  role?: string;
+  customType?: string;
+}>;
+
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: extensions/context-compat.ts [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+    const entries = compatible.buildContextEntries();
+    if (!Array.isArray(entries)) {
+      throw new Error(
+        "Unsupported session manager: buildContextEntries() returned no entries array",
+      );
+    }
+    return entries;
+  }
+
+  throw new Error(
```
LaansDole: Fail-closed throw: if a future runtime lacks both methods (or this function is ever called against a session manager mid-refactor), the extension dies at `session_start` — the exact failure mode this PR fixes. Failing open (`return []`) means `rulesAreInContext()` is false and the ruleset is simply re-injected, which is benign and self-healing. Deliberate choice or a gap?

extensions/context-compat.ts
```diff
@@ -0,0 +1,66 @@
...
+export function contextMessages(
+  sessionManager: unknown,
+): readonly ContextMessageMarker[] {
+  if (sessionManager === null || typeof sessionManager !== "object") {
+    throw new Error("Unsupported session manager: expected an object");
+  }
+
+  const compatible = sessionManager as CompatibleSessionManager;
+
+  if (typeof compatible.buildSessionContext === "function") {
```
LaansDole: `buildSessionContext()` is called without a guard — if it throws (session storage error, transient failure), the extension crashes at startup instead of degrading. Consider try/catch with the same fail-open fallback as above.

LaansDole (commented):
### Verified working
- `bun scripts/check_context_compat.ts` — passes (both shapes, marker ordering, non-marker messages ignored).
- `python3 -B -m unittest discover -s tests` — 15/15 pass.
- `python3 -B scripts/check_pi_extension.py --runtime omp` — passes against the real omp binary in an isolated RPC session (`--adhd` → `i-have-adhd` command registered, `ADHD ON` status).
- Adapter semantics are sound: `buildSessionContext().messages` is the resolved model context, so preferring it over `buildContextEntries()` is safe on upstream Pi too (both exist there; the scan keys on `customType`, and both shapes carry it).
- The `omp` install/update/uninstall docs and the manifest regression test are a nice touch.

### Suggestions (non-blocking)
1. **Fail-closed vs fail-open** (inline, `context-compat.ts:41`): the unsupported-manager throw reintroduces the exact crash class this PR fixes if a future runtime lacks both methods — and `buildSessionContext()` itself is called unguarded (inline, `context-compat.ts:21`), so a storage error there also kills the extension at session start. Failing open (`return []` → rules not in context → ruleset re-injected, which is benign and self-healing) is worth considering; if fail-closed is deliberate, a comment explaining why would help future readers.
2. **Overlap with #116** — that PR fixes the same bug with a minimal in-place shim. This one is strictly more complete (adapter, tests for both runtimes, docs).

ayghri:
Hardened context detection to fail open on unsupported APIs or context errors, allowing safe rules re-injection instead of startup crashes. Added regression coverage; all 15 tests pass.

Issue #118, opened 2026-08-16
docs(install): Hermes section says /i-have-adhd is a slash command, but Hermes has none


LaansDole:
**Problem**

INSTALL.md's Hermes section tells users to "Type `/i-have-adhd`" and claims the skill "is exposed as a slash command at the next session start". Hermes (Nous Research agent framework) has no per-skill slash-command mechanism — a user following the docs types `/i-have-adhd` and nothing happens.

**Root cause**

Hermes skills load differently from Claude Code / Pi: there is no skill-slash-command registry. A skill is activated by explicit user request (e.g. "use i-have-adhd mode") or loaded by the agent when the conversation matches the skill's description. The install command in the docs is correct; only the activation instruction is wrong.

**Repro**

1. `hermes skills install ayghri/i-have-adhd/skills/i-have-adhd` — works, skill lands in `~/.hermes/skills/`
2. Type `/i-have-adhd` in a session — unrecognized command; nothing activates

**Expected**

Docs describe the real activation path: ask the agent directly, or rely on description-driven loading; "stop adhd mode" turns it off.

PR #119, opened 2026-08-16
docs(install): fix Hermes activation instructions

LaansDole:
**Problem**

The Hermes section of INSTALL.md tells users to "Type `/i-have-adhd`" and says the skill is "exposed as a slash command at the next session start". An earlier revision of this PR claimed that instruction was wrong — that Hermes has no per-skill slash-command registry. That claim was **incorrect**: Hermes does register every installed skill as a slash command (see the review discussion on #119).

**Fix**

Keep the slash-command instruction and make the activation path precise: the skill is registered as a slash command at the next session start, `/reload-skills` picks it up in an already-running session, and natural language ("use i-have-adhd mode") works as a secondary path. `stop adhd mode` turns it off.

**Verification**

- `website/docs/user-guide/cli.md` § "Skill Slash Commands" and `website/docs/guides/work-with-skills.md`: every installed skill in `~/.hermes/skills/` is registered as a slash command automatically
- `get_skill_commands()` (`agent/skill_commands.py`) returns 159 skill commands on a real install including `/i-have-adhd`; `hermes chat -q "/i-have-adhd <task>"` expands the skill and runs it
- Clean-`HERMES_HOME` repro of the exact command in this file: `hermes skills install ayghri/i-have-adhd/skills/i-have-adhd` registers exactly one skill command, `/i-have-adhd`
- No older-version caveat applies: dynamic skill slash commands landed 2026-02-28 (commit `8e0c48e6d2`) and are present in the earliest tagged release (`v0.2.0` / `v2026.3.12`)
- `disable-model-invocation: true` in the skill's frontmatter means explicit invocation is the sanctioned path, which is what the file now says

Supersedes the premise in #118.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -298,7 +298,7 @@ Exceptions: explain fully when asked to explain. Confirm before destructive acti
 hermes skills install ayghri/i-have-adhd/skills/i-have-adhd
 ```
 
-Type `/i-have-adhd`. The skill installs into `~/.hermes/skills/` and is exposed as a slash command at the next session start.
+The skill installs into `~/.hermes/skills/` and is auto-discovered at the next session start. Hermes has no `/i-have-adhd` slash command (unlike Claude Code or Pi) — activate it by asking the agent directly (e.g. "use i-have-adhd mode"), or it may load itself when the conversation matches its description. Turn it off with "stop adhd mode".
 
 Prefer to browse first? Add this repo as a skill source (a "tap"), then search and install:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -298,7 +298,7 @@ Exceptions: explain fully when asked to explain. Confirm before destructive acti
 hermes skills install ayghri/i-have-adhd/skills/i-have-adhd
 ```
 
-The skill installs into `~/.hermes/skills/` and is auto-discovered at the next session start. Hermes has no `/i-have-adhd` slash command (unlike Claude Code or Pi) — activate it by asking the agent directly (e.g. "use i-have-adhd mode"), or it may load itself when the conversation matches its description. Turn it off with "stop adhd mode".
+Type `/i-have-adhd`. Hermes registers every installed skill as a slash command, so the skill name is the command; the skill installs into `~/.hermes/skills/` and is available from the next session start (`/reload-skills` picks it up in the current session without a restart). You can also activate it in natural language ("use i-have-adhd mode"). It stays on until you say "stop adhd mode".
 
 Prefer to browse first? Add this repo as a skill source (a "tap"), then search and install:
 


ayghri (changes requested):
Thanks for investigating the Hermes activation behavior. I cannot approve this change as written because its key premise conflicts with the current official Hermes documentation:

- https://github.com/NousResearch/hermes-agent/blob/main/website/docs/guides/work-with-skills.md : installed skills are automatically available as slash commands.
- https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/cli.md  : every installed skill in ~/.hermes/skills/ is registered as a slash command.

Please revise this PR rather than removing the /i-have-adhd activation path. The docs can retain the slash-command instructions, mention that newly installed skills take effect in a new session, and optionally note that natural-language activation is also supported. (if this PR targets a specific older Hermes version with different behavior, please identify that version or commit and document the distinction.)

kavix (approved):

LaansDole:
You're right, and thanks for the citations — I've revised the PR rather than removing the path (commit 6fcf628).

**What I got wrong.** I concluded "no skill slash-command mechanism exists" from the in-session `/help` list, which only enumerates the built-in command registry — dynamic skill commands aren't in it. They do exist.

**Re-verified against current Hermes:**

- `website/docs/user-guide/cli.md` § "Skill Slash Commands": every installed skill in `~/.hermes/skills/` is registered automatically, and the skill name becomes the command.
- `website/docs/guides/work-with-skills.md`: same, plus the tip that installed skills take effect in new sessions (`/reset`, or `--now`).
- Runtime: `get_skill_commands()` (`agent/skill_commands.py`) returns 159 skill commands on a real install, `/i-have-adhd` among them — and `hermes chat -q "/i-have-adhd <task>"` expands the skill and answers.
- Clean-`HERMES_HOME` repro of the exact install command in INSTALL.md: `hermes skills install ayghri/i-have-adhd/skills/i-have-adhd` into an empty home registers exactly one command, `/i-have-adhd`.

**Version question:** no distinction to document. Dynamic skill slash commands landed 2026-02-28 (commit `8e0c48e6d2`, "feat(skills): implement dynamic skill slash commands for CLI and gateway") and are present in the earliest tagged release, `v0.2.0` / `v2026.3.12`. Every released Hermes has it, so the file carries no version floor.

**One further correction.** The paragraph also said the skill "may load itself when the conversation matches its description". That contradicts the skill's own frontmatter (`disable-model-invocation: true`) and its description ("Invoke with /i-have-adhd"); Hermes doesn't implement that key at all. Natural language is now documented as a secondary path, not the primary one.

Current text:

> Type `/i-have-adhd`. Hermes registers every installed skill as a slash command, so the skill name is the command; the skill installs into `~/.hermes/skills/` and is available from the next session start (`/reload-skills` picks it up in the current session without a restart). You can also activate it in natural language ("use i-have-adhd mode"). It stays on until you say "stop adhd mode".

`/reload-skills` re-scans `~/.hermes/skills/` mid-session, so users who install while a session is open don't have to restart — say the word if you'd rather drop that clause and keep the paragraph minimal. Same for #118: I can correct its premise too, or close it as superseded.

PR #130, opened 2026-08-29
fix(opencode): universal fallback for 1.17/1.18 + opencode2 beta

mohammeddanishrabbani:
Fixes load failures on sst/opencode 1.18.25 and anomalyco/opencode2 0.0.0-beta-18600.

**Before:** `import { Plugin } from '@opencode-ai/plugin'` → `Export named 'Plugin' not found` on 1.18 (top-level now only re-exports tool) and `Unexpected export` due to conditional `export default`.

**After:** universal fallback: try `import { define } from '@opencode-ai/plugin/v2/promise'` (1.18/beta) → fallback to top-level `Plugin.define` (1.17) → final legacy v1 export (`config` + `experimental.chat.system.transform`). Single `export default` with ternary, no top-level await.

Also:
- fixes duplicate `.opencode/.opencode/plugins` path when `plugins[]` is in `.opencode/opencode.json` (resolve relative to that dir)
- add `.opencode/commands/i-have-adhd.md` (plural) for beta while keeping `.opencode/command/` for 1.17 compat
- tested: `opencode2 plugin list` now shows `i-have-adhd (active)` vs previous `(failed)`

### DIFF
--- a/.opencode/commands/i-have-adhd.md
+++ b/.opencode/commands/i-have-adhd.md
@@ -1,0 +1,5 @@
+---
+description: Shape output for a reader with ADHD for the rest of this session
+---
+
+Use the `i-have-adhd` skill and apply its ruleset to every response for the rest of this session: lead with the next action, number multi-step work, restate state across turns, suppress tangents, give concrete time estimates, and make wins visible. These rules persist until I say "stop adhd mode" or "normal mode".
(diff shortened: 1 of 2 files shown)


ayghri:
Thanks for investigating this. We’re keeping support scoped to standard OpenCode, not forks or beta releases.

The proposed fallback cannot work reliably: @opencode-ai/plugin/v2/promise is a static import, so a runtime without that entry point fails before any fallback can run. I reproduced this on OpenCode 1.18.18.

Closing this PR rather than adding untested compatibility code. Feel free to update and reopen it once OpenCode 2 is released.

PR #146, opened 2026-09-08
Fix #132: import OpenCode plugin via pathToFileURL for Windows

### LINKED ISSUE #132: OpenCode plugin test driver cannot import Windows absolute paths
## Reproduction

On Windows with Python 3.13 and Node 24.14.1, run:

`python -m unittest discover -s tests -v`

The three `tests/test_opencode_plugin.py` cases fail before their assertions. Node reports `ERR_UNSUPPORTED_ESM_URL_SCHEME`: absolute paths on Windows must be valid `file://` URLs, but `tests/opencode_plugin_driver.mjs` calls `await import(pluginPath)` with a `C:\...` path.

## Expected

The existing synthetic OpenCode-plugin tests should pass on Windows as they do on POSIX. This affects only the test driver; it does not change the skill or plugin behavior.

## Suggested direction

Convert the argv path with Node `pathToFileURL(...).href` from `node:url` before dynamically importing it. The existing three test cases provide regression coverage.

## Triage

At the report baseline `cbe69fb83c08a37cf54d5ec9ec6bb88c8bc9973c`, I rechecked open and closed issues/PRs plus discussions for this error and the test-driver wording. I found no duplicate or prior fix. Open PR #130 changes the runtime OpenCode plugin fallback and does not modify the test driver.

Validation used only temporary test fixtures and local Node/Python execution; no model, external service, account, or personal data was used.

FekyBaz:
Fixes #132. On Windows, dynamically importing the plugin with a raw absolute path fails with ERR_UNSUPPORTED_ESM_URL_SCHEME (received protocol 'c:'). Convert the argv path with pathToFileURL(...).href from node:url before importing so it becomes a valid file:// URL. Test-driver only; no skill or plugin behavior change. Verified: tests/test_opencode_plugin.py 3/3 pass on Windows (were 3/3 fail before). Full suite: only the 4 pre-existing test_judge/test_run_evals env errors remain (also fail on clean tree).

### DIFF
--- a/tests/opencode_plugin_driver.mjs
+++ b/tests/opencode_plugin_driver.mjs
@@ -2,8 +2,10 @@
 // `experimental.chat.system.transform` hook against an empty system prompt, and
 // prints the resulting system text so tests can assert on the injected banner.
 // Nothing is printed when the hook injects nothing (always-on flag absent).
+import { pathToFileURL } from 'node:url';
+
 const pluginPath = process.argv[2];
-const { default: init } = await import(pluginPath);
+const { default: init } = await import(pathToFileURL(pluginPath).href);
 const hooks = await init();
 const output = { system: [] };
 await hooks['experimental.chat.system.transform']({}, output);


rajpratham1 (approved):
This is a focused fix for Windows plugin imports. Using pathToFileURL(pluginPath).href correctly converts the filesystem path into a URL before the dynamic import(), addressing the reported Windows compatibility issue. I don't see any blocking issues in this PR.

ayghri:
Thanks for the PR. merged as is.

Issue #129, opened 2026-08-25
SessionStart always-on hook silently no-ops when CLAUDE_PLUGIN_ROOT is not exported (Claude Code desktop app, macOS)


swhite1122:
### Summary

On the Claude Code **desktop app** (macOS, Claude Code 2.1.241), the always-on `SessionStart` hook silently does nothing. No banner, no error, exit 0. ADHD mode never loads and there is no signal that anything went wrong.

The shipped hook command is:

```json
"command": "node -e \"(async()=>{const root=process.env.CLAUDE_PLUGIN_ROOT||process.env.PLUGIN_ROOT;if(root)await import(...)})().catch(()=>{})\""
```

The desktop app **substitutes** `${CLAUDE_PLUGIN_ROOT}` into hook `args`, but does not **export** `CLAUDE_PLUGIN_ROOT` into the hook process environment. So `root` is `undefined`, the `if (root)` guard skips the import, and the process exits 0 having done nothing.

This appears to be a regression introduced by the fix for #70: moving from `sh "${CLAUDE_PLUGIN_ROOT}/hooks/always-on.sh"` (template substitution) to reading the env var at runtime. #70 at least failed loudly; this fails silently.

### Evidence

Same machine, same install, same `~/.claude/.i-have-adhd-always` flag present, plugin `enabled=true`.

**Terminal `claude`** — works:

```
[DEBUG] Loading hooks from plugin: i-have-adhd
[DEBUG] Hook SessionStart:startup (SessionStart) success:
ADHD MODE ACTIVE (always-on). ...
```

**Desktop app**, same session start type (`SessionStart:clear`) — no banner. Another plugin's SessionStart hook in the same session fired normally, so hooks themselves are working; that plugin uses `"args": ["${CLAUDE_PLUGIN_ROOT}/hooks/activate.js"]` (substitution) rather than reading the env var.

Reproducing the no-op directly:

```console
$ env -u CLAUDE_PLUGIN_ROOT -u PLUGIN_ROOT node -e "(async()=>{const root=process.env.CLAUDE_PLUGIN_ROOT||process.env.PLUGIN_ROOT;if(root)await import(...)})().catch(()=>{})"
$ echo $?
0
```

No output, no error.

### Why this is worse than a crash

[...]

Issue #91, opened 2026-07-31
hook codex Support


SpxZhu:
Currently, SessionStart Hooks only work properly within Claude Code. Can a universal Hook be provided that can also work in Codex?

ayghri:
Do you have any diagnostic for the failure/lack of support on codex? I haven't tried the hook on codex yet.

lwu-st:
Here's the fix fyi:
i-have-adhd used args with ${CLAUDE_PLUGIN_ROOT}, and Codex passed that arg literally, so node tried to load Users/dev/${CLAUDE_PLUGIN_ROOT}/hooks/always-on.mjs and exited 1.

fanqiecd:
Thanks for opening this. I reproduced the failure on Windows with Codex using the current `main` hook manifest (`i-have-adhd` v0.1.0) and Node v26.5.0.

### Observed

Codex reports:

```text
hook exited with code 1
```

`main/hooks/hooks.json` currently contains:

```json
"command": "node",
"args": ["${CLAUDE_PLUGIN_ROOT}/hooks/always-on.mjs"]
```

The Codex hook runner does not handle this Claude-specific `args` / `${CLAUDE_PLUGIN_ROOT}` combination the same way Claude Code does. In the observed path, only `node` is executed and the SessionStart JSON on stdin is parsed as JavaScript.

Minimal PowerShell reproduction:

```powershell
$payload = '{"session_id":"codex-repro","cwd":"C:\\Temp","hook_event_name":"SessionStart"}'
$payload | node
```

This exits 1:

```text
SyntaxError: Unexpected token ':'
ExitCode=1
```

If the argument is forwarded literally instead, the same manifest fails with a `MODULE_NOT_FOUND` path containing `${CLAUDE_PLUGIN_ROOT}`.

### Expected

The `SessionStart` hook should invoke `hooks/always-on.mjs` and exit 0 (silently when the opt-in flag is absent) on Codex/Windows as well as Claude Code.

### Minimal Codex-side fix

Putting the complete command in `command` and using Codex's plugin-root placeholder fixes the failure:

```diff
           {
             "type": "command",
-            "command": "node",
-            "args": ["${CLAUDE_PLUGIN_ROOT}/hooks/always-on.mjs"],
+            "command": "node \"${PLUGIN_ROOT}/hooks/always-on.mjs\"",
             "timeout": 5,
             "statusMessage": "Checking i-have-adhd always-on flag..."
           }
```

[...]

PR #150, opened 2026-09-08
fix(hooks): warn on stderr when plugin root is unresolved (#129)

wakqasahmed:
## Summary

On the Claude Code desktop app, `CLAUDE_PLUGIN_ROOT`/`PLUGIN_ROOT` never reach the SessionStart hook's process environment (#129), so the always-on hook did nothing at all — no banner, no error, exit 0. A user with always-on enabled had no way to tell it from always-on simply being off.

Full architectural fixes (resolving the plugin root from the hook script's own location, or switching to host-side `${CLAUDE_PLUGIN_ROOT}` substitution in `args`) would need real desktop-app testing I don't have access to here, and the existing test suite locks in the current single-`command`-string shape specifically because that's what unbroke Codex in #91 — reintroducing `args` risks that regression again without a way to verify it against Codex either. So this takes the reporter's own explicitly-stated fallback instead: the hook now checks the opt-in flag first (independent of plugin root — it's a plain file-existence check under `CLAUDE_CONFIG_DIR`) and only writes a one-line stderr diagnostic when a user who *has* opted in also hits the missing-root case. Everyone who hasn't opted into always-on sees no behavior change at all.

- [ ] **Human-authored**
- [x] **Autonomous agent-authored**
- [ ] **Hybrid**

**Agent/tool and model/version:** Claude Code, Sonnet 5 (claude-sonnet-5)

**Agent contribution:** Wrote the fix and tests, and manually exercised all four flag/root combinations against the real hook command in a shell before writing them up as tests.

**Human verification:** The repo owner (wakqasahmed) directed this fix and will review the diff. Not verified against the real desktop app itself — only against the hook command run standalone with the desktop app's reported environment shape (no `CLAUDE_PLUGIN_ROOT`/`PLUGIN_ROOT`, opt-in flag present/absent).

**Known limitations or uncertain results:** This doesn't make always-on actually work on the desktop app — it only makes the failure visible instead of silent, which is the reporter's own stated "at minimum" ask. Making it actually work needs a maintainer decision on #129's fuller options (a real fallback-resolution mechanism or a second chained hook entry), plus desktop-app access to verify.

## Labels

**Target label:** Target:Integrations

**Author label:** Author:AI

**Workflow labels:** bug

## Safety and side effects

**Side effects, permissions, network access, and cost:** None. Adds one `fs.existsSync` check (same flag-file check `always-on.mjs` already does) and one conditional `stderr.write`; still exits 0 in every case.

## Compatibility

- [ ] This is a breaking change

**Migration or rollback notes:** None. Users who never opted into always-on see byte-identical behavior (verified by test and by hand). Opted-in users only gain a stderr line in the one case that was previously silently broken.

## Verification

- `python3 -m unittest discover -s tests -v` — 43 tests, OK (2 new: silent when not opted in and root missing; stderr diagnostic when opted in and root missing)
- `python3 scripts/run_evals.py validate` — cases valid
- `git diff --check` — clean
- Manually ran the exact hook command from `hooks/hooks.json` through all 4 flag×root combinations in a shell before writing the corresponding tests; output matched what's asserted (silent / banner / silent / stderr diagnostic).

## Final accountability

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node -e \"(async()=>{const root=process.env.CLAUDE_PLUGIN_ROOT||process.env.PLUGIN_ROOT;if(root)await import(require('node:url').pathToFileURL(require('node:path').join(root,'hooks','always-on.mjs')).href)})().catch(()=>{})\"",
+            "command": "node -e \"(async()=>{const fs=require('node:fs'),os=require('node:os'),path=require('node:path');const cfgDir=process.env.CLAUDE_CONFIG_DIR||path.join(os.homedir(),'.claude');const flagged=fs.existsSync(path.join(cfgDir,'.i-have-adhd-always'));const root=process.env.CLAUDE_PLUGIN_ROOT||process.env.PLUGIN_ROOT;if(root){await import(require('node:url').pathToFileURL(path.join(root,'hooks','always-on.mjs')).href)}else if(flagged){process.stderr.write('i-have-adhd: always-on is enabled but CLAUDE_PLUGIN_ROOT/PLUGIN_ROOT was not provided to the SessionStart hook, so the ruleset was not injected this session (see https://github.com/ayghri/i-have-adhd/issues/129).\\n')}})().catch(()=>{})\"",
             "timeout": 30,
             "statusMessage": "Checking i-have-adhd always-on flag..."
           }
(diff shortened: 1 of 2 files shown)


wakqasahmed:
Hi @ayghri — noticed this PR doesn't have a reviewer assigned yet — it's been sitting for a while, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look, or point me to who should? Thanks!

PR #152, opened 2026-09-09
fix(evals): handle non-object judge responses (#151)

### LINKED ISSUE #151: Non-object judge responses stop the entire batch
`judge.py` stops the whole batch if the judge returns valid JSON with the wrong shape, such as `null` or `{"A": null, "B": {}}`.

Reproduced on `24d22f7`, Python 3.12.10:

```python
import sys
sys.path.insert(0, "scripts")
import judge

judge.parse_judge_scores(
    '{"A": null, "B": {}}',
    ("direct-answer", 1),
    {"baseline": "A", "candidate": "B"},
)
# AttributeError: 'NoneType' object has no attribute 'get'
```

The retry path catches `ValueError`, but these inputs raise `TypeError` or `AttributeError`. A two-group test with a stubbed judge confirmed that the second group is never processed.

Checking that the document and each verdict are objects would let these failures use the existing retry/skip behavior. I'll send a focused fix with regression tests. This validation gap remains after #97.

AI-authored with Codex (GPT-6): investigation, local reproduction, and drafting. The submitter requested this work but has not independently rerun the tests. No paid model calls.

tkv00:
## Summary

Fixes #151.

Valid JSON such as `null` or `{"A": null}` can raise an uncaught `TypeError` or `AttributeError` in `parse_judge_scores()`, stopping the whole batch.

Validate the response and each verdict as objects, raising a contextual `ValueError` so the existing retry and skip handling can continue with later groups. Add four regression tests for invalid shapes, retry recovery, and batch continuation.

- [ ] **Human-authored** — substantive implementation and text were produced by a human.
- [x] **Autonomous agent-authored** — an agent planned and produced most of the substantive change.

**Agent/tool and model/version:** OpenAI Codex, GPT-6.

**Agent contribution:** Bug reproduction, implementation, tests, and issue/PR text.

**Human verification:** The submitter selected the bug and requested submission. Full diff review and independent test execution are unconfirmed; draft pending human review.

**Known limitations or uncertain results:** Four existing Windows tests fail because `sh` is unavailable (`WinError 2`), as also reported in #146.

## Labels

**Target label:** Target:Evals

**Author label:** Author:AI

**Workflow labels:** bug

Requested labels; upstream label permissions are unavailable.

## Safety and side effects

**Side effects, permissions, network access, and cost:** No new dependencies or permissions. Tests use temporary files and mocks; no network or provider costs.

## Compatibility

- [x] Canonical and mirrored skill files are synchronized when applicable.

**Migration or rollback notes:** None. Valid verdicts retain the same output format. Skill files, manifests, and installation instructions are unchanged.

## Verification

Windows, Python 3.12.10 embedded. `python` below denotes the interpreter invoked by its full path.

- `python -B -m unittest discover -s tests -p test_judge.py -k non_object -v` — four new tests fail before the fix and pass afterward.
- `python -B -m unittest discover -s tests -v` — 41 passed, four pre-existing `sh`-dependent errors, 45 total.
- `python -B scripts/run_evals.py validate` — passed.
- `git diff --check` — passed.

**Behavior evals:** Not applicable; skill rules are unchanged. No paid evaluations were run.

## Final accountability

### DIFF
--- a/scripts/judge.py
+++ b/scripts/judge.py
@@ -118,12 +118,16 @@ def parse_judge_scores(
     """Turn one judge response into score rows keyed by condition, not label."""
     case_id, trial = group_key
     verdicts = json.loads(_strip_code_fence(payload))
+    if not isinstance(verdicts, dict):
+        raise ValueError(f"{case_id}/trial {trial}: judge response must be a JSON object")
     rows = []
     for condition, label in sorted(labels.items()):
         where = f"{case_id}/trial {trial}/label {label}"
         if label not in verdicts:
             raise ValueError(f"{where}: judge returned no verdict for label {label}")
         verdict = verdicts[label]
+        if not isinstance(verdict, dict):
+            raise ValueError(f"{where}: verdict must be a JSON object")
         row: dict[str, Any] = {"case_id": case_id, "trial": trial, "condition": condition}
         for dimension in DIMENSIONS:
             value = verdict.get(dimension)
--- a/tests/test_judge.py
+++ b/tests/test_judge.py
@@ -3,6 +3,7 @@ import sys
 import tempfile
 import unittest
 from pathlib import Path
+from unittest.mock import patch
 
 
 ROOT = Path(__file__).resolve().parents[1]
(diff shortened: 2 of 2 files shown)


kavix (approved):

Anurag852001:
W

Anurag852001 (approved):

scripts/judge.py
```diff
@@ -118,12 +118,16 @@ def parse_judge_scores(
     """Turn one judge response into score rows keyed by condition, not label."""
     case_id, trial = group_key
     verdicts = json.loads(_strip_code_fence(payload))
+    if not isinstance(verdicts, dict):
```
Anurag852001, line 121: Can you explain why you added this?

Anurag852001 (changes requested):

scripts/judge.py
tkv00: This is defensive validation rather than a fix for a specific `null` response I observed. The judge is only prompted to return an object; that shape is not enforced, and `json.loads()` also accepts `null`, arrays, and scalar values. If one of those gets through, the code fails later with a `TypeError` or `AttributeError`, bypassing the existing `ValueError` retry path and stopping the batch. Raising `ValueError` here lets the current retry/skip flow handle it. I added tests for the two outcomes we care about: a retry can recover, and an exhausted group does not prevent later groups from being judged.

PR #158, opened 2026-09-10
fix(hooks): ship hooks.codex.json so Codex session start gets a JSON envelope

### LINKED ISSUE #156: SessionStart hook output is plain text; Codex rejects it as invalid JSON and drops the ruleset
Follow-up to #91, which is now fixed: `${CLAUDE_PLUGIN_ROOT}` resolves and the hook no longer exits 1. The hook runs — but Codex still discards it, for a different reason.

## Observed

`i-have-adhd` 0.3.0, `codex-cli` 0.154.0, Node v25.2.1, macOS.

Every Codex session opens with:

```text
• Hook failed
  └ hook returned invalid session start JSON output
```

The hook exits 0, but writes plain text to stdout:

```console
$ R=~/.codex/plugins/cache/i-have-adhd/i-have-adhd/0.3.0
$ printf '{"hook_event_name":"SessionStart","cwd":"/tmp","session_id":"t","source":"startup"}' \
    | CLAUDE_PLUGIN_ROOT="$R" node "$R/hooks/always-on.mjs"
ADHD MODE ACTIVE (always-on). The ruleset below applies to every response. "stop adhd mode" turns...
```

Claude Code accepts plain stdout from a `SessionStart` hook and treats it as context. Codex does not — it requires a JSON envelope, and on anything else it drops **the entire hook output**. The ruleset never reaches the model.

The failure is silent in the way that matters: the error line doesn't name which hook failed, and the plugin looks installed and enabled. On a machine with several plugins it is not obvious that ADHD mode simply isn't on in Codex.

## Expected

Either output shape is accepted by Codex:

- empty stdout, or
- a single JSON object:

```json
{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"ADHD MODE ACTIVE ..."}}
```

## Suggested fix: ship `hooks/hooks.codex.json`

Codex reads a plugin's `hooks/hooks.codex.json` in preference to `hooks/hooks.json` when present. That is the mechanism `remember@claude-plugins-official` (0.29.1) uses, and it works — its Codex session_start hook returns cleanly on the same machine where this one fails.

That keeps the Claude Code path untouched: `hooks.json` keeps emitting plain text, and a Codex-only entry emits the envelope. A thin wrapper around the existing `always-on.mjs` is enough — no change to the ruleset itself.

## Verified workaround (not a fix)

For anyone hitting this before it lands: register a wrapper in your own `~/.codex/hooks.json` that runs the plugin hook and wraps its stdout.

```js
// envelope.mjs <command> [args...] — runs the command, wraps plain-text stdout in the envelope.
// Empty stdout stays empty; stdout that is already JSON passes through untouched.
```

Confirmed working: the ruleset lands in the session as a `role=developer` item, verified in
`~/.codex/sessions/rollout-*.jsonl`.

Two caveats. The plugin's own hook still runs alongside it, so the `Hook failed` line stays. And if Codex's automatic sync of hooks/skills/plugins is enabled, it removes the wrapper: that sync mirrors the Claude Code configuration into Codex, and a Codex-only entry has no counterpart to mirror, so it is dropped on the next sync. Turning that sync off — keeping only chats and projects synced — makes the wrapper persist.

*Edit: an earlier version of this issue blamed a Codex-run cleanup of "non-working hooks" for removing the wrapper. That was wrong; the sync is the cause.*

Happy to open a PR with the `hooks.codex.json` if that shape is welcome.

STiFLeR7:
Fixes #156.

## What changed and why
Codex requires SessionStart hook output to be either empty stdout or a single JSON envelope. It rejects the plain-text output that always-on.mjs writes, which Claude Code accepts directly, and silently drops the entire ruleset with no indication of which hook failed.

Codex reads hooks/hooks.codex.json in preference to hooks/hooks.json when present, the same mechanism the issue points to in remember@claude-plugins-official. Added hooks/codex-session-start.mjs, a thin wrapper that runs the existing always-on.mjs unchanged, captures its stdout, and re-emits it as a hookSpecificOutput envelope when non-empty. hooks.json and always-on.mjs are untouched, so Claude Code keeps receiving plain text exactly as before.

## Observable behavior before and after
Before: a Codex session with the always-on flag set shows "Hook failed: hook returned invalid session start JSON output" and the ruleset never reaches the model.
After: the same session receives a valid hookSpecificOutput envelope containing the same ruleset text, with no hook failure.

## Safety and compatibility
No change to always-on.mjs, hooks.json, or the sh/PowerShell fallbacks. The new wrapper only intercepts stdout for the duration of the import and restores it immediately after, matching the existing opt-in/fail-safe/no-network behavior of the hook it wraps. If always-on.mjs exits early with no output (no opt-in flag, missing skill file, or an internal error), the wrapper also emits nothing, preserving the existing silent no-op behavior.

## Verification performed
Ran the full hooks test suite locally (Node available, sh/PowerShell not on this machine): python -m pytest tests/test_always_on_hooks.py -v, 8 passed. Updated run_codex_hook to read hooks.codex.json instead of hooks.json, added test_codex_command_wraps_the_hook_output_in_the_session_start_envelope asserting the parsed JSON envelope shape and hookEventName, and added test_codex_hook_uses_a_launcher alongside the renamed test_claude_hook_uses_a_launcher so both launcher shapes are covered independently. The two existing silent-output tests (no opt-in flag, missing plugin) needed no changes since the wrapper already emits nothing when always-on.mjs produces no text.

## Authorship and provenance
Autonomous agent-authored. I read the issue, the existing hooks.json launcher pattern, and always-on.mjs, then implemented and tested the fix with no human writing code. A human reviewed the diff and the test run before this PR was opened. No failed checks to report; the full existing test file plus the new/updated cases all pass locally.

### DIFF
--- a/hooks/codex-session-start.mjs
+++ b/hooks/codex-session-start.mjs
@@ -1,0 +1,31 @@
+// Codex-only SessionStart wrapper. Codex requires SessionStart hooks to emit
+// either empty stdout or a single JSON envelope, unlike Claude Code, which
+// accepts always-on.mjs's plain-text output directly. This captures that
+// plain text and re-emits it wrapped, without changing always-on.mjs itself.
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+
+const root = process.env.CLAUDE_PLUGIN_ROOT || process.env.PLUGIN_ROOT;
+if (!root) process.exit(0);
+
+const chunks = [];
+const originalWrite = process.stdout.write.bind(process.stdout);
+process.stdout.write = (chunk) => {
+  chunks.push(chunk);
+  return true;
+};
+
+try {
+  await import(pathToFileURL(path.join(root, "hooks", "always-on.mjs")).href);
+} finally {
+  process.stdout.write = originalWrite;
+}
+
+const text = chunks.join("");
+if (text) {
+  process.stdout.write(
+    JSON.stringify({
+      hookSpecificOutput: { hookEventName: "SessionStart", additionalContext: text },
+    }),
+  );
+}
--- a/hooks/hooks.codex.json
+++ b/hooks/hooks.codex.json
@@ -1,0 +1,17 @@
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node -e \"(async()=>{const root=process.env.CLAUDE_PLUGIN_ROOT||process.env.PLUGIN_ROOT;if(root)await import(require('node:url').pathToFileURL(require('node:path').join(root,'hooks','codex-session-start.mjs')).href)})().catch(()=>{})\"",
+            "timeout": 30,
+            "statusMessage": "Checking i-have-adhd always-on flag..."
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/tests/test_always_on_hooks.py
+++ b/tests/test_always_on_hooks.py
@@ -54,7 +54,7 @@ class AlwaysOnHookTest(unittest.TestCase):
         )
 
     def run_codex_hook(self, plugin_root=None):
-        config = json.loads((ROOT / "hooks" / "hooks.json").read_text())
+        config = json.loads((ROOT / "hooks" / "hooks.codex.json").read_text())
         hook = config["hooks"]["SessionStart"][0]["hooks"][0]
         env = os.environ.copy()
         env["CLAUDE_CONFIG_DIR"] = str(self.config_dir)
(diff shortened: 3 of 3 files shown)


ayghri:
I was not able to reproduce this issue. Official documentation (https://learn.chatgpt.com/docs/hooks#sessionstart) says plain-text output is accepted.

STiFLeR7:
Checked the docs you linked: they do confirm plain text on stdout is accepted for SessionStart, which contradicts what I assumed. I can't explain the reporter's exact error (codex-cli 0.154.0, 'invalid session start JSON output') from that alone, maybe a version-specific behavior or another installed hook's output getting merged with this one. Could use the reporter's help narrowing that down before deciding whether this PR is still needed.

PR #182, opened 2026-09-11
fix(evals): reject NaN evaluation budgets

### LINKED ISSUE #181: --allow-unmetered turns off the budget check
So I was running evals with a Codex runner and passed --budget-usd 5 --allow-unmetered, thinking the budget would still hold things back. It didn't. The thing just ran every case × trial without stopping and printed Reported cost: $0.0000 at the end. Kind of a problem if someone does this against a paid API.

The budget logic looks at reported_cost, which only goes up when the runner returns a dollar amount:

reported_cost += float(cost or 0)
But with --allow-unmetered, _parse_response always returns cost = None. So reported_cost stays at zero, remaining never hits zero, and this check:

remaining = args.budget_usd - reported_cost
if remaining <= 0:
    return 2

never fires. The flag basically makes --budget-usd decorative.

kudala-bharani:
## Summary

`--budget-usd nan` passes the current range check: both `nan <= 0` and `nan > 25` are false. That lets an invalid budget reach the runner and makes the remaining-budget comparisons ineffective.

Require the budget to satisfy `0 < budget <= 25` instead. NaN is then rejected before any runner call or output-directory creation. Valid budgets and the existing error message stay unchanged.

The regression test exercises nine CLI inputs, including NaN, infinities, out-of-range values, and valid budgets of $0.01 and $25. It also checks that valid budgets are still forwarded to the runner. This does not change `--allow-unmetered` or claim to resolve #181.

- [ ] **Human-authored**
- [x] **Autonomous agent-authored**
- [ ] **Hybrid**

**Agent/tool and model/version:** OpenAI Codex; exact model/version identifier is not available in this session.

**Agent contribution:** Found and reproduced the validation gap, implemented the fix and regression test, ran the checks below, and prepared this PR. A separate Codex agent reviewed the diff and ran the evaluation-harness tests.

**Human verification:** I reviewed the complete two-file diff and authorized submission. The tests below were run by Codex; I am not claiming separate human test execution.

**Known limitations or uncertain results:** Tested on macOS with Python 3.14.6 and Node 25.9.0, not Windows or Linux. Hook tests exercised Node and sh; PowerShell was unavailable. No live provider calls were made.

## Labels

**Target label:** Target:Evals

**Author label:** Author:AI

**Workflow labels:** bug

## Safety and side effects

**Side effects, permissions, network access, and cost:** No new dependencies, network calls, or configuration changes. The new test uses a mocked runner and temporary files, with no provider cost.

## Compatibility

Canonical and mirrored skills, platform manifests, hooks, and installation documentation are unchanged. No synchronization or integration changes are needed.

**Migration or rollback notes:** No migration required. Only a budget outside the existing valid range is newly rejected.

## Verification

- `python3 -m unittest discover -s tests -v` — 52 tests passed, with `TMPDIR=/private/tmp/adhd-contrib.hWDPkB` to avoid macOS temporary-path aliases.
- `python3 -m unittest discover -s tests -p test_run_evals.py -k budget_range -v` — passed all nine budget subtests. Before the fix, both NaN inputs failed with `ValueError not raised`.
- `python3 scripts/run_evals.py validate` — evaluation cases valid.
- `git diff --check main...HEAD` — clean.

**Behavior evals:** Not run: this changes input validation in the evaluation harness, not the response rules. Pi/OMP and Claude CLI validation were not run because no integrations changed.

## Final accountability

### DIFF
--- a/scripts/run_evals.py
+++ b/scripts/run_evals.py
@@ -268,7 +268,7 @@ def run_evaluations(args: argparse.Namespace) -> int:
         if row.get("condition") == args.condition and row.get("runner") == args.runner
     )
 
-    if args.budget_usd <= 0 or args.budget_usd > 25:
+    if not 0 < args.budget_usd <= 25:
         raise ValueError("--budget-usd must be greater than 0 and no more than 25")
 
     args.output.parent.mkdir(parents=True, exist_ok=True)
--- a/tests/test_run_evals.py
+++ b/tests/test_run_evals.py
@@ -274,6 +274,52 @@ class EvaluationHarnessTest(unittest.TestCase):
             self.assertEqual(0, run_evals.run_evaluations(args))
             self.assertTrue(marker.exists())
 
+    def test_budget_range_is_enforced_before_runner_invocation(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp_path = Path(tmp)
+            runner_config = tmp_path / "runners.json"
+            runner_config.write_text(
+                json.dumps({
+                    "stub": {
+                        "command": ["stub-runner"],
+                        "response_format": "claude-json",
+                        "budget_flag": "--max-budget-usd",
+                    }
+                }),
+                encoding="utf-8",
+            )
+            completed = subprocess.CompletedProcess(
+                args=["stub-runner"], returncode=0,
+                stdout=json.dumps({"result": "102", "total_cost_usd": 0.01}),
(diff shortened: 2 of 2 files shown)


ayghri:
These edge cases do not affect the skill and are purely on the eval side, for which the dev should be able to handle on themselves. Striving for verifiably correct code in a skill repo is too ambitious of a goal.

PR #191, opened 2026-09-11
docs: add Traditional Chinese (zh-TW) README and INSTALL

Franky100-pig:
Adds a zh-TW (繁體中文) translation alongside the existing zh-CN one, and wires the new language into the switcher in README.md and every translated README so the language list stays consistent across all eight languages.

The text was produced with OpenCC (s2twp: Simplified to Traditional with Taiwan phrase conversion) from the zh-CN source and then reviewed. Code blocks, commands, paths, and URLs are byte-identical to the source. Three vocabulary cases were kept against the OpenCC phrase suggestion because the current form is the Taiwan standard: 官方文件 (documentation, not 檔案 which means archive) and 游標 (cursor).

The switcher entry uses the text label 繁體中文 instead of a flag emoji, since Traditional Chinese is a script used across several regions rather than one country. Easy to swap if a flag is preferred.

Verification: python3 -m unittest tests.test_install_docs passes, so INSTALL.zh-TW.md satisfies the Zed path checks that run over every INSTALL.*.md. The full suite shows 5 pre-existing macOS failures in OpenCodePluginTest (the /var to /private/var symlink issue), unrelated to this change and addressed in a separate PR; this change adds no new failures.

- [ ] **Human-authored** — substantive implementation and text were produced by a human.
- [ ] **Autonomous agent-authored** — an agent planned and produced most of the substantive change.

**Agent contribution:** 

**Human verification:** 

## Labels

**Target label:** 

**Author label:** 

**Workflow labels:** 

## Safety and side effects

- [ ] The change does not access or expose secrets, private files, or unrelated user/repository data.
- [ ] Scripts, hooks, workflows, and evals are bounded and do not create surprising or irreversible side effects.
- [ ] No destructive, privileged, production, externally visible, or persistent action occurs without explicit user intent and appropriate safeguards.
- [ ] Network access, third-party code, permissions, and provider costs are minimized and documented.
- [ ] Prompt text, examples, and fixtures contain no hidden instructions that weaken safety or expand agent authority.

## Compatibility

- [ ] This is not a breaking change.
- [ ] Relevant platform manifests and installation documentation were reviewed.

## Verification

- `<command>` — `<result>`

**Behavior evals:** 

## Final accountability

- [ ] All failed, skipped, or unrun checks are disclosed above.

### DIFF
--- a/.github/install/INSTALL.zh-TW.md
+++ b/.github/install/INSTALL.zh-TW.md
@@ -1,0 +1,716 @@
+# 安裝方法
+
+<details>
+<summary><strong>Antigravity (<code>agy</code>)</strong></summary>
+
+### 安裝
+
+```bash
+agy plugin install https://github.com/ayghri/i-have-adhd
+```
+
+### 驗證
+
+```bash
+agy plugin list
+```
+
+### 更新
+
+```bash
+agy plugin uninstall i-have-adhd
+agy plugin install https://github.com/ayghri/i-have-adhd
+```
+
+### 解除安裝
+
+```bash
+agy plugin uninstall i-have-adhd
+```
+
+也可以保留安裝並將其關閉：`agy plugin disable i-have-adhd`。
+
+### 始終啟用（可選）
+
+新增到 `~/.gemini/GEMINI.md`：
+
+```markdown
+## 輸出風格
+
+讀者有 ADHD。請讓每條回覆都便於立即執行：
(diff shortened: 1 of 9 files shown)


ayghri:
Personally, I prefer traditional characters. But this will mostly add translation debt if we take into consideration that simplified CN translation should be more than enough.

PR #203, opened 2026-09-13
fix(hooks): trim PowerShell trailing blank lines;Fix/powershell trailing newlines

aprashnasuraj-dev:

- [x] **Human-authored** — substantive implementation and text were produced by a human.
- [ ] **Autonomous agent-authored** — an agent planned and produced most of the substantive change.

**Agent contribution:** 

**Human verification:** 

## Labels

**Target label:** 

**Author label:** 

**Workflow labels:** 

## Safety and side effects

- [ ] The change does not access or expose secrets, private files, or unrelated user/repository data.
- [ ] Scripts, hooks, workflows, and evals are bounded and do not create surprising or irreversible side effects.
- [ ] No destructive, privileged, production, externally visible, or persistent action occurs without explicit user intent and appropriate safeguards.
- [ ] Network access, third-party code, permissions, and provider costs are minimized and documented.
- [ ] Prompt text, examples, and fixtures contain no hidden instructions that weaken safety or expand agent authority.

## Compatibility

- [ ] This is not a breaking change.
- [ ] Relevant platform manifests and installation documentation were reviewed.

## Verification

- `<command>` — `<result>`

**Behavior evals:** 

## Final accountability

- [ ] All failed, skipped, or unrun checks are disclosed above.

### DIFF
--- a/hooks/always-on.ps1
+++ b/hooks/always-on.ps1
@@ -40,6 +40,7 @@ try {
   } else {
     ""
   }
+  $body = $body.TrimEnd([char]13, [char]10)
 
   $banner = 'ADHD MODE ACTIVE (always-on). The ruleset below applies to every response. ' +
     '"stop adhd mode" turns it off for this session; delete '
--- a/tests/test_always_on_hooks.py
+++ b/tests/test_always_on_hooks.py
@@ -81,7 +81,7 @@ class AlwaysOnHookTest(unittest.TestCase):
     @staticmethod
     def normalize(stdout):
         # The banner embeds the flag path. On Windows the sh runtime joins it
-        # with "/" while node and PowerShell join with "\"; both name the same
+        # with "/" while node and PowerShell join with "\\"; both name the same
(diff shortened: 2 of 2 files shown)


aprashnasuraj-dev:
This looks like more than simply “GitHub being down.” There are now reproducible cases showing transient runner→STS network stalls being amplified by the action’s timeout behavior. In #1915, an STS request can remain pending indefinitely because the retry wrapper only runs after a promise rejects; a stalled request never reaches that retry path. #1848 also shows v6 ETIMEDOUT retries lasting long enough for the OIDC JWT itself to expire.

So I’d describe this as intermittent network/routing trouble combined with an action-side resilience gap, rather than a global STS outage. Until the per-request behavior is fixed, I would bound both the workflow/job timeout and action-timeout-s, keep retries finite, and preserve the failed runner/IP/STS endpoint information when it happens.

In other words: the initial stall may be infrastructure/network-related, but an hours-long hang does not have to be.

PR #219, opened 2026-09-16
fix(hooks): emit the ruleset as UTF-8 from the PowerShell hook

SurefireStudios:
## Summary

`hooks/always-on.ps1` writes the ruleset with `[Console]::Out.Write`, which encodes through the console code page — IBM437 on a default en-US Windows PowerShell.

Every non-ASCII character in `SKILL.md` is destroyed before the agent sees it: `Regla: di la conclusion primero. 中文规则。 Café — règle.` arrives as `Regla: di la conclusion primero. ????? Caf? - r?gle.`

Node and sh both emit UTF-8, so this is the PowerShell fallback alone.

The loss is entirely on output: dumping code points between read and write gives `67,97,102,233,32,20013,25991` from `File::ReadAllLines` but `b'Caf\x82 ??'` from stdout.

The fix writes UTF-8 bytes straight to the standard output stream rather than assigning `[Console]::OutputEncoding`, which is process-global and can throw in a redirected host.

This matters because the ruleset is injected verbatim into every response, so a runtime that re-encodes it corrupts the rules themselves.

The parity suite could not catch this twice over: every fixture is ASCII, and `run_hook` passed `text=True` with no `encoding`, so stdout was decoded through the locale too and mojibake would have been compared against mojibake.

Both are fixed, and the new case is appended at the end of the class, clear of #173 and #203.

- [ ] **Human-authored**
- [x] **Autonomous agent-authored** — an agent planned and produced most of the substantive change.
- [ ] **Hybrid**

**Agent/tool and model/version:** Claude Code, Opus 5.

**Agent contribution:** Investigation, reproduction, root-cause isolation, the fix, the test, and this description.

**Human verification:** The submitting human requested the work and reviewed the diff, but has **not** independently re-run the suite. Every number here comes from the agent's runs.

**Known limitations or uncertain results:** All local runs were on a single Windows 11 / en-US host; the corruption depends on the console code page, so other locales lose a different set of characters. Independently reproduced green on a GitHub `windows-latest` runner (below).

## Labels

**Target label:** Target:Integrations
**Author label:** Author:AI
**Workflow labels:** bug

## Safety and side effects

**Side effects, permissions, network access, and cost:** None. Output-path change only — no new dependency, no network, no filesystem write, and notably not `[Console]::OutputEncoding`. The `try`/`catch` and `exit 0` contract is untouched, so the hook still cannot block session start. `UTF8Encoding $false` writes no BOM.

## Compatibility

- [x] Canonical and mirrored skill files are synchronized when applicable — n/a, no skill file changed.
- [x] Relevant platform manifests and installation documentation were reviewed — unaffected.

**Migration or rollback notes:** ASCII output is byte-for-byte identical, so no existing user's behavior changes. Both `New-Object System.Text.UTF8Encoding` and `[Console]::OpenStandardOutput()` work on Windows PowerShell 5.1 and PowerShell 7+.

## Verification

- `python -m pytest tests/test_always_on_hooks.py -q` — **before** `1 failed, 7 passed` (the PowerShell subTest of the new case), **after** `8 passed`
- `python -m pytest tests -q` — **before** `3 failed, 67 passed`, **after** `3 failed, 68 passed`; those three are `tests/test_judge.py`, pre-existing on Windows and unrelated (#159 has them)
- This repo's `plugin-load-check.yml` run unmodified on a fork — `hook-parity (windows-latest)` and `hook-parity (ubuntu-latest)` both `Ran 8 tests / OK`, with `test_runtimes_emit_the_body_as_utf8 ... ok` on each. A second, independent Windows host matters here because the bug depends on the code page.

I also diffed all three runtimes byte-for-byte across nine `SKILL.md` shapes (ASCII, CJK, accented Latin, emoji, CRLF, no trailing newline, empty body, `---` in body, BOM): four diverged before, one after. The remaining one is a UTF-8 BOM — PowerShell strips it, Node and sh pass it through. PowerShell is arguably right there, so which behavior wins is a maintainer call rather than an obvious bug, and I left it out to keep this a one-line output fix. Happy to file it separately.

**Behavior evals:** Not run — no skill-behavior change; this PR only affects how existing bytes are encoded on the way out.

## Final accountability


SurefireStudios:
Upstream Actions don't run on fork PRs, so I ran this repo's own `plugin-load-check.yml` unmodified on a fork to exercise the `hook-parity` matrix — the job that exists for exactly this kind of change.

**Both legs green, including a second, independent Windows host:**

```
success   hook-parity (windows-latest)      Ran 8 tests   OK
success   hook-parity (ubuntu-latest)       Ran 8 tests   OK
```

The new case ran on both:

```
test_runtimes_emit_the_body_as_utf8 ... ok        (windows-latest)
test_runtimes_emit_the_body_as_utf8 ... ok        (ubuntu-latest)
```

That matters more than my local run does. The bug depends on the console code page, so a pass on a GitHub Windows runner — a different host with a different default from mine — is the evidence that the fix is not specific to my machine. The ubuntu leg confirms the Node and sh paths are untouched.

